Issue #850, opened 2015-06-03
编译1.0.15时提示Some files do not have the e xpected license header 


dxxfire:
JDK版本：build 1.6.0_25-b06
构建环境：win7 64位Maven3.2.5
F:\it\druid>mvn clean install -Dmaven.test.skip=true
[INFO] Scanning for projects...
[INFO]
[INFO] ------------------------------------------------------------------------
[INFO] Building druid 1.0.15-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[WARNING] The POM for com.oracle:ojdbc6:jar:11.2.0.3 is missing, no dependency i
nformation available
[WARNING] The POM for proxool:proxool:jar:0.9.1 is missing, no dependency inform
ation available
[WARNING] The POM for proxool:proxool-cglib:jar:0.9.1 is missing, no dependency
information available
[INFO]
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ druid ---
[INFO] Deleting F:\it\druid\target
[INFO]
[INFO] --- buildnumber-maven-plugin:1.1:create (default) @ druid ---
[INFO] Storing buildNumber: 2015-06-03 11:43:46 at timestamp: 1433303026790
[INFO] Executing: cmd.exe /X /C "git rev-parse --verify HEAD"
[INFO] Working directory: F:\it\druid
[INFO] Storing buildScmBranch: UNKNOWN_BRANCH
[INFO]
[INFO] --- maven-resources-plugin:2.6:resources (default-resources) @ druid ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] Copying 58 resources
[INFO]
[INFO] --- maven-compiler-plugin:3.0:compile (default-compile) @ druid ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 877 source files to F:\it\druid\target\classes
[WARNING] 注意：某些输入文件使用或覆盖了已过时的 API。
[WARNING] 注意：要了解详细信息，请使用 -Xlint:deprecation 重新编译。
[WARNING] 注意：F:\it\druid\src\main\java\com\alibaba\druid\support\console\Drui
dStat.java 使用了未经检查或不安全的操作。
[WARNING] 注意：要了解详细信息，请使用 -Xlint:unchecked 重新编译。
[INFO]
[INFO] --- maven-resources-plugin:2.6:testResources (default-testResources) @ dr
uid ---
[INFO] Not copying test resources
[INFO]
[...]

PR #854, opened 2015-06-05
fixed for license check

wenshao:


wenshao:
for https://github.com/alibaba/druid/issues/850

PR #855, opened 2015-06-05
bug fixed for pg sqlparser

wenshao:


wenshao:
https://github.com/alibaba/druid/issues/853

PR #857, opened 2015-06-07
improved mysql parser support alter table add multicolumns

wenshao:


wenshao:
https://github.com/alibaba/druid/issues/849

Issue #881, opened 2015-06-29
Hive 配置连接池报错


xiaohui886688:
java.sql.SQLException: Method not supported
	at org.apache.hadoop.hive.jdbc.HiveConnection.getHoldability(HiveConnection.java:354)
	at com.alibaba.druid.filter.FilterChainImpl.connection_getHoldability(FilterChainImpl.java:325)
	at com.alibaba.druid.filter.FilterAdapter.connection_getHoldability(FilterAdapter.java:861)
	at com.alibaba.druid.filter.FilterChainImpl.connection_getHoldability(FilterChainImpl.java:322)
	at com.alibaba.druid.proxy.jdbc.ConnectionProxyImpl.getHoldability(ConnectionProxyImpl.java:244)
	at com.alibaba.druid.pool.DruidConnectionHolder.<init>(DruidConnectionHolder.java:86)
	at com.alibaba.druid.pool.DruidDataSource.init(DruidDataSource.java:631)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)

wenshao:
已经修复，等下一版本吧

flzx3qc:
请问是哪个版本后修复了这个问题呢

PR #888, opened 2015-07-05
improved hive-jdbc support

wenshao:


wenshao:
https://github.com/alibaba/druid/issues/881

PR #999, opened 2015-11-30
DruidDataSource.getStatDataForMBean() should use getActiveCount for A…

laogao:
DruidDataSource.getStatDataForMBean() should use getActiveCount for ActiveCount, just as getStatData() does.


wenshao:
感谢你的反馈

wenshao:
感谢你的反馈

PR #1036, opened 2016-01-12
mysql sql parser support long byte datatype

wenshao:


wenshao:
https://github.com/alibaba/druid/issues/1035

PR #1062, opened 2016-01-26
PagerUtils support count(distinct)

wenshao:
https://github.com/alibaba/druid/issues/1060


wenshao:
bug已经修复，等druid 1.0.17吧

PR #1079, opened 2016-02-17
bug fixed for issue 1076

wenshao:


wenshao:
https://github.com/alibaba/druid/issues/1076

Issue #1155, opened 2016-03-31
druid与atomikos整合，不支持jdbc:mysql:replication写法？


insist211314:
`
java.lang.ClassCastException: com.sun.proxy.$Proxy30 cannot be cast to com.mysql.jdbc.ConnectionImpl
	at com.alibaba.druid.util.MySqlUtils.createXAConnection(MySqlUtils.java:30)
	at com.alibaba.druid.pool.xa.DruidXADataSource.createPhysicalXAConnection(DruidXADataSource.java:72)
	at com.alibaba.druid.pool.xa.DruidXADataSource.getXAConnection(DruidXADataSource.java:48)
	at com.atomikos.datasource.xa.jdbc.JdbcTransactionalResource.createXAConnection(JdbcTransactionalResource.java:167)
	at com.atomikos.datasource.xa.jdbc.JdbcTransactionalResource.refreshXAConnection(JdbcTransactionalResource.java:100)
	at com.atomikos.datasource.xa.XATransactionalResource.getXAResource(XATransactionalResource.java:316)
	at com.atomikos.datasource.xa.XATransactionalResource.recover(XATransactionalResource.java:450)
	at com.atomikos.datasource.xa.XATransactionalResource.setRecoveryService(XATransactionalResource.java:418)
	at com.atomikos.icatch.config.Configuration.notifyAfterInit(Configuration.java:465)
	at com.atomikos.icatch.config.Configuration.init(Configuration.java:449)
	at com.atomikos.icatch.jta.UserTransactionManager.startupTransactionService(UserTransactionManager.java:71)
	at com.atomikos.icatch.jta.UserTransactionManager.initializeTransactionManagerSingleton(UserTransactionManager.java:60)
	at com.atomikos.icatch.jta.UserTransactionManager.checkSetup(UserTransactionManager.java:52)
	at com.atomikos.icatch.jta.UserTransactionManager.init(UserTransactionManager.java:122)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
[...]

PR #1156, opened 2016-04-02
bug fixed for mysql url jdbc:mysql:replicate

wenshao:


wenshao:
https://github.com/alibaba/druid/issues/1155

PR #1177, opened 2016-04-21
Enhancement for Teradata select and insert clauses

usakey:
This PR mainly focuses on bug fix and new tokens added to TD **Select** and **Insert** clause.

1. bug fix for `SchemaStatVisitor` as column will return `null` if key exists whilst value is null;
1. add support for `SEL`, `MOD` and `QUALIFY`;
1. add support for `SQLSubqueryTableSource` for TD;
1. add `aliasQueryMap` to store mapping from alias to expression, as there may exist `null` in alias map. This feature mainly enables us to track the mapping between **real target columns** and **real source columns**.


wenshao:
能否先签署CLA，这是公司开源的要求，贡献代码需要放弃版权

PR #1214, opened 2016-05-06
db2 sql parser support concat binary operator

wenshao:


wenshao:
https://github.com/alibaba/druid/issues/1190

PR #1286, opened 2016-06-13
mysql解析后去掉标识符的反引号 #1267

gyk001:


gyk001:
just ignore it, my understanding is wrong 😭 . 
get column name shoud use class com.alibaba.druid.stat.TableStat.Column which has right name

PR #2056, opened 2017-10-20
 #2055 change the function to parse the sql in the right of '~'

sunsun314:
Reason:  
  We find is not right to parse the sql ‘ ~ (id = 1)' & ‘ ~ id = 1'
Type :  
  Bug fix
Influences：
  Change the function to parse the sql right behind the '~'


wenshao:
能否补上testcase，避免以后问题重现出现

PR #2287, opened 2018-01-25
add default query timeout for mssql

jun1st:
add default query timeout for mssql

druid version 1.1.6 fails to create jdbc connection to database due to this issue


wenshao:
已修改，新的实现是如果<0就忽略设置queryTimeout。

PR #2810, opened 2018-09-18
解决问题#2809，com.alibaba.druid.spring.boot.autoconfigure.DruidDataSourceWrapper类无public修饰

hedeejay:
…SourceWrapper： a non-visible class，not public class

org.springframework.aop.framework.AopConfigException: Could not generate CGLIB subclass of class [class com.alibaba.druid.spring.boot.autoconfigure.DruidDataSourceWrapper]: Common causes of this problem include using a final class or a non-visible class; nested exception is org.springframework.cglib.core.CodeGenerationException: java.lang.reflect.InvocationTargetException-->null


hotdust:
Because of this problem, opentracing-spring-cloud-jdbc-starter can not be used with druid starter.
It's appreciated that merge fix to master if no other problem.

lizongbo:
现在已经有public修饰符了

PR #2896, opened 2018-11-05
fix issue https://github.com/alibaba/druid/issues/2859

lonegunmanb:
The problem is caused by setting IdentifierExpr's ResolvedTableSource to a SQLExprTableSource while SQLExprTableSource's expr is THE IdentifierExpr it self, then if we enter SQLExprTableSource.findTableSourceWithColumn, it'll cause an endless recursion call. This is a simple patch to prevent an IdentifierExpr point an TableSource into itself.


lizongbo:
最新版druid运行正常,,因此关闭

PR #3032, opened 2019-01-21
modify OracleOutputVisitor for insert multiple records

TaeRoen:
For example, I have a oracle sql statement, 
**"INSERT INTO user (id, name) VALUES ('1', ?),('2', ?)"**
when use SQLUtils to format it, it output 
**"INSERT INTO user (id, name) VALUES ('1', ?)"**
OracleOutputVisitor will discard second ValuesClause,

I referenced PGOutputVisitor 298 line and modify OracleOutputVisitor
PGOutputVisitor 298 line:
```
if (x.getValues() != null) {
     println();
     print0(ucase ? "VALUES " : "values ");
     printlnAndAccept(x.getValuesList(), ", ");
} else {
    if (x.getQuery() != null) {
        println();
        x.getQuery().accept(this);
    }
}
```


TaeRoen:
Even though oracle is not support this sql, druid will hide this error.

PR #3035, opened 2019-01-23
validationQueryTimeout in OracleValidConnectionChecker

cutesnail:
issues: The oracle driver will call cancel when the query timeout, but if the timeout is caused by executing in a closed connection(closed by database), the cancel command would not work and the checker throws runtime exception until the socket read timeout(config by oracle.jdbc.ReadTimeout, default 600000 milliseconds).
So the time used to check the connection may be longer than the validationQueryTimeout configuration and will cause upper invoker's reading time out.

fix: Execute the connection check by submitting a asynchronous task to ExecutorService and retrieve the result from the Future instance.


Yelijah:
I meet this problem too, but why this commit didn't be merged?

cutesnail:
> 
> 
> I meet this problem too, but why this commit didn't be merged?

Try other ojdbc driver.

HSHunterR:
oracle version 11.2  which ojdbc version should be choose？ can I use ojdbc6  11.2.0.3.0 solve this problem？

Issue #3036, opened 2019-01-23
Replace toArray(new T[size]) with toArray(new T[0])


hc-codersatlas:
Hi, noticed many occurrences of toArray that could be made faster with a find and replace.

Performance: replace toArray(new T[size]) with toArray(new T[0])

https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion (also contains benchmarks).

Could provide some simple performance improvements thanks!

wenshao:
for ArrayList, list.toArray(new String[list.size()]) has better performance than list.toArray(new String[0])

hc-codersatlas:
Hi, did you read the linked article? It explicitly explains why new T[0] is faster than new T[size].

PR #3037, opened 2019-01-23
Optimize toArray (#3036)

hc-codersatlas:
replace toArray(new T[size]) with toArray(new T[0]) for better performance


wenshao:
Your suggestion is incorrect.

PR #3130, opened 2019-03-16
Apply p3c pmd plugin/rules and fix all violations of priority 1

blindpirate:
This PR applies https://github.com/alibaba/p3c pmd plugin/rules and fixes all violations of priority 1.


lizongbo:
冲突太多，请以最新master代码更新，再重新提交pr

PR #3379, opened 2019-08-04
simple verify syntax tool jar

yangmx5:
build jar for sql syntax verify
used:  java -jar xxx.jar {$sql_path}


wenshao:
请不要做格式化，不好做diff

PR #3500, opened 2019-10-02
improve pg statement parser. fix the missing bracket in select block

Duncan-tree-zhou:
when a select block is bracket. PGStatementParser will miss the bracket when giving the output.

fix the issue

test case class is :
com.alibaba.druid.postgresql.PGUnionTest


wenshao:
已支持

PR #4019, opened 2020-11-03
fix hive  TBLPROPERTIES parse.   eg:tblproperties (parquet.compression = 'snappy' ) .without ''

thorntree:
fix hive  TBLPROPERTIES parse.   eg:tblproperties (parquet.compression = 'snappy' ) .without ''


thorntree (commented):
fix hive TBLPROPERTIES parse. eg:tblproperties (parquet.compression = 'snappy' ) .without ''

PR #4031, opened 2020-11-10
parse alter table problem

yoloz:
1,rename to miss space  
2,alter constraint condition reversed


wenshao:
https://github.com/alibaba/druid/releases/tag/1.2.4
请用新版本

PR #4243, opened 2021-04-19
DruidConnectionHolder: Support setter/getter of default transaction isolation level

fuyufjh:
Add setter/getter to `defaultTransactionIsolation` in `DruidConnectionHolder`.

In our case, the default transaction isolation level may be differed in each session, and we always call 'setTransactionIsolation' after getting a connection from pool. In this scenario Druid always send a `SET SESSION TRANSACTION ISOLATION LEVEL` command to backend database, which downgrades the performance significantly. 

We hope this behavior can be suppressed by allowing us set current isolation level to `defaultTransactionIsolation`, so that during putting back the connection (aka. `reset()`) the druid will preserve the current isolation-level, until next time the connection is reused and another isolation level would be set.

### DIFF
--- a/src/main/java/com/alibaba/druid/pool/DruidAbstractDataSource.java
+++ b/src/main/java/com/alibaba/druid/pool/DruidAbstractDataSource.java
@@ -1388,7 +1388,7 @@ public abstract class DruidAbstractDataSource extends WrapperAdapter implements 
         }
 
         if (validConnectionChecker != null) {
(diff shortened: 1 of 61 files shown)


zrlw:
it's might better that adding a configurable parameter (say keepUnderlyingTransactionIsolation), if the parameter is true, then the underlyingTransactionIsolation will not be reset at DruidConnectionHolder#reset.

PR #4276, opened 2021-05-13
avoid NPE of MySqlCreateTableParser

yanhuqing666:
avoid NPE of MySqlCreateTableParser


lizongbo:
冲突太多，请重新提交新的pr

lizongbo:
冲突太多，请重新提交新的pr

PR #4328, opened 2021-06-21
JdbcSqlStat.addResultSetHoldTimeNano() should not convert to nanos

moonfruit:
For this:

https://github.com/alibaba/druid/blob/584dcb6c69224cc9e31200230d9977c75c67b004/src/main/java/com/alibaba/druid/stat/JdbcSqlStat.java#L1057-L1062

When calling `executeAndResultHoldTimeHistogramRecord()`, it should not convert value to nanos by `(statementExecuteNano + resultHoldTimeNano) / 1000 / 1000`. Because `executeAndResultHoldTimeHistogramRecord()` expect nanos:

https://github.com/alibaba/druid/blob/584dcb6c69224cc9e31200230d9977c75c67b004/src/main/java/com/alibaba/druid/stat/JdbcSqlStat.java#L715-L735


moonfruit:
@wenshao Is there any problem with this?

PR #4398, opened 2021-07-30
Fix StatViewFilter

lfz757077613:
When I use StatViewFilter like below
image
And it throws an exception if I enter http://localhost:8080/druid
image
Perhaps, its better to delete this code? I think it is useless
```
        if (requestURI.equals(servletPath)) {
            httpResp.sendRedirect(httpReq.getRequestURI() + '/');
        }
```


zrlw:
the confilcts could be resolved as this pr is based on old incompatible version.
commit a new PR #5613 based on this PR.

PR #4399, opened 2021-07-30
StatViewFilter support reset enable

lfz757077613:
StatViewFilter don't support to set resetEnable as StatViewServlet, why?


lizongbo:
代码重复太多，请重新提交新的pr

PR #4489, opened 2021-10-01
issue-4190 fix PR. added more UTF8 lowercase and uppercase characters that can be the first letter of an identifier

angbisoft:
… 1st letter of an identifier. For details see https://en.wikipedia.org/wiki/List_of_Unicode_characters

### DIFF
--- a/src/main/java/com/alibaba/druid/sql/parser/CharTypes.java
+++ b/src/main/java/com/alibaba/druid/sql/parser/CharTypes.java
@@ -43,9 +43,7 @@ public class CharTypes {
     private final static boolean[] firstIdentifierFlags = new boolean[256];
     static {
         for (char c = 0; c < firstIdentifierFlags.length; ++c) {
-            if (c >= 'A' && c <= 'Z') {
-                firstIdentifierFlags[c] = true;
-            } else if (c >= 'a' && c <= 'z') {
+            if ((c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') || (c >= 'À' && c <= 'ÿ')) {
                 firstIdentifierFlags[c] = true;
             }
         }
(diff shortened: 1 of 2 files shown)


wenshao:
please signed CLA

zrlw:
https://dev.mysql.com/doc/refman/8.0/en/identifiers.html
```
Internally, identifiers are converted to and are stored as Unicode (UTF-8). The permissible Unicode characters in identifiers are those in the Basic Multilingual Plane (BMP). Supplementary characters are not permitted. Identifiers thus may contain these characters:
Permitted characters in unquoted identifiers:
ASCII: [0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore)
Extended: U+0080 .. U+FFFF
Permitted characters in quoted identifiers include the full Unicode Basic Multilingual Plane (BMP), except U+0000:
ASCII: U+0001 .. U+007F
Extended: U+0080 .. U+FFFF
ASCII NUL (U+0000) and supplementary characters (U+10000 and higher) are not permitted in quoted or unquoted identifiers.
Identifiers may begin with a digit but unless quoted may not consist solely of digits.
Database, table, and column names cannot end with space characters.
Beginning with MySQL 8.0.32, use of the dollar sign as the first character in the unquoted name of a database, table, view, column, stored program, or alias is deprecated and produces a warning. This includes such names used with qualifiers (see [Section 9.2.2, “Identifier Qualifiers”](https://dev.mysql.com/doc/refman/8.0/en/identifier-qualifiers.html)). The dollar sign can still be used as the leading character of such an identifier when it is quoted according to the rules given later in this section.
The identifier quote character is the backtick (`):
mysql> SELECT * FROM `select` WHERE `select`.id > 100;
If the [ANSI_QUOTES](https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_ansi_quotes) SQL mode is enabled, it is also permissible to quote identifiers within double quotation marks:
mysql> CREATE TABLE "test" (col INT);
ERROR 1064: You have an error in your SQL syntax...
mysql> SET sql_mode='ANSI_QUOTES';
mysql> CREATE TABLE "test" (col INT);
Query OK, 0 rows affected (0.00 sec)
```
as mentioned above, the first letter of an identifier could be:
```
ASCII: [0-9,a-z,A-Z$_] (basic Latin letters, digits 0-9, dollar, underscore)
Extended: U+0080 .. U+FFFF
identifier quote character: backtick (`)
if ANSI_QUOTES is enabled: double quotation marks(")
```

zrlw:
if taking ```U+0080 .. U+FFFF``` as first letter of an identifier, they should be checked at isFirstIdentifierChar() instead of putting them in firstIdentifierFlags as the length of firstIdentifierFlags array is only 256.

zrlw:
i see, you want add only Latin capital and small letters which value less than 255 to firstIdentifierFlags array.
the decimal values of latin capital and small letters range between 192('À') and 255('ÿ'),  but multiplication sign '×'(215) and division_sign '÷' (247) should be excluded.

zrlw:
this PR is based on old incompatible version, the branch conflicts could be resolved by merging.
submit a new PR #5604 based on this PR.

PR #4574, opened 2021-11-15
fix flaky test in MySqlCreateExternalCatalogTest1

CharlesZKQ:
In com/alibaba/druid/bvt/sql/mysql/create/MySqlCreateExternalCatalogTest1.java, there are two tests, both of them are flaky due to the non-deterministic property of HashMap, which is used by SQLStatement.toString(). The toString() function called com.alibaba.druid.sql.dialect.mysql.visitor.MySqlOutputVisitor.visit(), which used the iterator of HashMap. I fixed these two tests by generating all  possible combinations of strings that could be output by toString() function. More specifically, I put all properties into a list and generate all permutation of the list to construct the string, which solved the flakiness of these two tests.


lizongbo:
代码冲突，不方便合并，欢迎基于最新代码重新提交pr

PR #4575, opened 2021-11-15
fix flaky test in MySqlCreateDatabaseTest.test_2

CharlesZKQ:
In com/alibaba/druid/bvt/sql/mysql/create/MySqlCreateDatabaseTest.java, test_2() is flaky due to the non-deterministic property of HashMap, proved in [document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html), it is used by SQLUtils.toMySqlString(). This function called com.alibaba.druid.sql.dialect.mysql.visitor.MySqlOutputVisitor.visit(), which used the iterator of HashMap. I fixed these two tests by generating all possible combinations of strings that could be output by toMySqlString() function. The method is similar to [pull4574](https://github.com/alibaba/druid/pull/4574).


lizongbo:
代码冲突了

PR #4576, opened 2021-11-15
fix flaky test in MySqlCreateExternalCatalogTest2

CharlesZKQ:
In com/alibaba/druid/bvt/sql/mysql/create/MySqlCreateExternalCatalogTest2.java, test_0() is flaky due to the non-deterministic property of HashMap, proved in proved in [document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html), it is used by SQLStatement.toString(). This function called com.alibaba.druid.sql.dialect.mysql.visitor.MySqlOutputVisitor.visit(), which used the iterator of HashMap. I fixed it by generating all possible combinations of strings that could be output by toString() function. The method is similar to [pull4574](https://github.com/alibaba/druid/pull/4574).


lizongbo:
代码冲突，不方便合并，欢迎基于最新代码重新提交pr

PR #4577, opened 2021-11-15
fix flaky tests in MySqlCreateResourceGroupTest

CharlesZKQ:
In com/alibaba/druid/bvt/sql/mysql/create/MySqlCreateResourceGroupTest.java, test_alter(), test_create(), test_create2()  are all flaky tests due to the non-deterministic property of HashMap, proved in [document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html), it is used by SQLUtils.toMySqlString(). This function called com.alibaba.druid.sql.dialect.mysql.visitor.MySqlOutputVisitor.visit(), which used the iterator of HashMap. I fixed these two tests by generating all possible combinations of strings that could be output by toMySqlString() function, with general function generateAllPossibleRes(). It takes two parameters, prefix and suffix, which could be used for all flaky tests in this file.


lizongbo:
代码冲突，不方便合并，欢迎基于最新代码重新提交pr

PR #4578, opened 2021-11-15
fix flaky test MySqlCreateResourceGroupTest_hive

CharlesZKQ:
In com/alibaba/druid/bvt/sql/mysql/create/MySqlCreateResourceGroupTest_hive.java, test_alter(), test_create() are all flaky tests due to the non-deterministic property of HashMap, proved in [document](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html), it is used by SQLUtils.toMySqlString(). This function called com.alibaba.druid.sql.dialect.mysql.visitor.MySqlOutputVisitor.visit(), which used the iterator of HashMap. I fixed these two tests by generating all possible combinations of strings that could be output by toMySqlString() function, with general function generateAllPossibleRes(). It takes one parameter prefix, which could be used for all flaky tests in this file.


kimmking:
good job, can you fix conflicting files first?

lizongbo:
代码冲突，不方便合并，欢迎基于最新代码重新提交pr

PR #4778, opened 2022-05-11
[FEAT] Extend H2OutputVisitor to make it could convert mysql scripts to h2, general available

AlphaHinex:
Could not parse and convert `DELIMITER` token yet.


wenshao:
代码冲突了，请修复重后重亲提交

PR #5009, opened 2022-11-17
mysql 默认的 validationQuery 添加 /* ping */ 前缀

vdiskg:


wenshao:
为什么要加这个前缀？

vdisk-group:
> 为什么要加这个前缀？

https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-usagenotes-j2ee-concepts-connection-pooling.html
根据 mysql 的官方文档，当 sql 语句以 `/* ping */` 开头时 jdbc  驱动会尝试发送 ping 来替代执行 sql，比直接执行 `SELECT 1` 更高效

> MySQL Connector/J can validate the connection by executing a lightweight ping against a server. 
In the case of load-balanced connections, this is performed against all active pooled internal connections that are retained. 
This is beneficial to Java applications using connection pools, as the pool can use this feature to validate connections. 
Depending on your connection pool and configuration, this validation can be carried out at different times:
> 1. Before the pool returns a connection to the application.
> 2. When the application returns a connection to the pool.
> 3. During periodic checks of idle connections.
> To use this feature, specify a validation query in your connection pool that starts with /* ping */. Note that the syntax must be exactly as specified.
 This will cause the driver send a ping to the server and return a dummy lightweight result set.
 When using a ReplicationConnection or LoadBalancedConnection, the ping will be sent across all active connections.

kimmking:
直接调用Connection.isValid(int timeout)，岂不更好（内部也是调用的ping）。

kimmking:
补充一下：isValid是JDBC4.0添加的方法，所以各种不同数据库驱动都有实现。比如在pgsql里，会执行一个空sql。

zrlw:
mysql实现的isValid方法是调用的pingInternal方法，对于LoadBalancedMySQLConnection、ReplicationMySQLConnection这种多重连接而言，此方法只能验证其中一条连接；而"/* ping */ xxx"这种调用的是connection的ping方法，多重连接的ping方法默认是遍历验证全部连接的。
所以mysql建议用"/* ping */ xxx"这种方式以支持各种类型的connections。

PR #5107, opened 2023-02-15
fix: CreateConnectionThread that keeps endlessly reconnecting even after the data source has been properly closed 

lvxiao1:
After calling the `DruidDataSource.close()` method, the interrupt status of the `CreateConnectionThread` is set to true. However, if the MySQL is configured with `autoReconnect=true`, it will trigger the MySQL reconnection mechanism, which includes a call to `Thread.sleep()`, resetting the interrupt status.

`ConnectionImpl:connectWithRetries()`
image

`CreateConnectionThread`：Due to the reset of the interrupt status, calling `lock.lockInterruptibly()` fails to throw an exception.
image

If the connection is always NULL due to constant exceptions, it will result in an infinite loop.
image

Here is the code for reproducing the issue：
```java
DruidDataSource dataSource = new DruidDataSource();
Properties properties = new Properties();
properties.put("druid.url", "jdbc:mysql://10.0.0.1:3306/xx?autoReconnect=true");
properties.put("druid.username", "root");
properties.put("druid.password", "root");
dataSource.configFromPropety(properties);
new Thread(() -> {
    try (DruidPooledConnection connection = dataSource.getConnection()) {

    } catch (Exception e) {
        e.printStackTrace();
    }
}).start();

Thread.sleep(1000);
dataSource.close();
System.out.println("close success");
LockSupport.park();
```

### DIFF
--- a/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
+++ b/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
@@ -2949,13 +2949,15 @@ public class DruidDataSource extends DruidAbstractDataSource implements DruidDat
                     continue;
                 }
 
-                boolean result = put(connection);
-                if (!result) {
-                    JdbcUtils.close(connection.getPhysicalConnection());
-                    LOG.info("put physical connection to pool failed.");
-                }
+                if (connection != null) {
+                    boolean result = put(connection);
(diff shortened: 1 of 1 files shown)


lvxiao1:
@wenshao please check

lizongbo:
有新代码实现了

PR #5117, opened 2023-02-20
fix: add new class load path

newboy2004:
fixbug：#5060  druid一直加载不到MysqlConnection，导致CPU 100%


kimmking (approved):

kimmking:
the above fix is not completed. and fix it again: https://github.com/alibaba/druid/commit/b9f98fc40b4bfb1a8df2772c36ebf9dc186c34a2

newboy2004:
ok,Will there be another version？

PR #5121, opened 2023-02-21
[fix] 连接池启用phyTimeoutMillis之后造成连接泄露

huangjinteng:
shrink中connections移除相关的holder
https://github.com/alibaba/druid/issues/5119


kimmking:
Actually, it's not thread-safe to remove from a list.
And we expect avoid connection-leak by closing them just in time, not by gc collection.

PR #5147, opened 2023-03-02
fix nonstandard message

gongxuanzhang:
current code : char array toString()  is  char[]@hashcode
we want get char array value rather than char hashcode.
update to  new String(char[])


kimmking:
You should sign CLA at first. @gongxuanzhang

PR #5196, opened 2023-04-01
Precompute function and rebuild the AST sqlstatement

LZH-Work-Station:
Usually, we hope that the elements in the SQL statement will be parsed and processed before they are actually submitted to the executor (such as checking the permissions of fields and the number of partitions, etc.), but some functions, including some UDFs defined by users, make it difficult to preprocess these elements. For example, the commonly used DATEADD function, if users need to check the partition range involved in the SQL query before submitting it to the executor, it will be quite difficult. In fastsql, we provide a good Function function to define functions and use SQLEvalVisitorUtil to calculate the results of functions and write them into functions, but we do not reconstruct the original SqlStatement (using the results of functions to replace the function expression). Here, I have made a complement to this.


zrlw:
git push提交使用的user.name和user.email和github账户不一致，签不了CLAassistant，关了这个PR重新提交吧，提交前查看配置的账户和邮箱，要与github账号一致。
$ git config --global -l

PR #5279, opened 2023-05-16
fix sql parsing for create statements without AS

craynic:
Fixes for parsing create statements like `create table some_table select * from another_table`.


wenshao:
这个是什么数据库的语法？

PR #5543, opened 2023-12-02
Fix shrink with on fatal error

zrlw:
see #5542
1. increase fatalErrorCountLastShrink at handleFatalError(), otherwise emptySignal will be called again at shrink().
2. remove onFatalError from  emptySignal judging condition of shrink(), otherwise emptySignal will be always called by shrink() when onFatalError is true.

### DIFF
--- a/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
+++ b/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
@@ -1990,6 +1990,8 @@ public class DruidDataSource extends DruidAbstractDataSource
             ReentrantLock dataSourceLock = holder.getDataSource().lock;
             dataSourceLock.lock();
             try {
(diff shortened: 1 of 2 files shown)


zrlw:
```
Error:  /home/runner/work/druid/druid/core/src/main/java/com/alibaba/druid/sql/visitor/SQLASTOutputVisitor.java:[1938,23] variable parameters is already defined in method printMethodParameters(com.alibaba.druid.sql.ast.expr.SQLMethodInvokeExpr,java.util.List<com.alibaba.druid.sql.ast.SQLExpr>)
```
it's so weird, 
1. this PR has nothing to with SQLASTOutputVisitor.java
2. SQLASTOutputVisitor.java is compiled ok in my local machine.
3. i don't find the method: printMethodParameters(com.alibaba.druid.sql.ast.expr.SQLMethodInvokeExpr,java.util.List<com.alibaba.druid.sql.ast.SQLExpr>)

zrlw:
oops, it's caused by branch confliction.

PR #5570, opened 2023-12-09
optimize shrink by moving retained connections during checking cycle and removing shrinkbuffer

zrlw:
see #5567
1. move retained connections forward by remaining position pointer during shrink checking cycle.
2. remove shrinkbuffer.
3. clean the old references of the retained connections that have been moved forward to the new positions by array copy.

this PR refers to #5569

### DIFF
--- a/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
+++ b/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
@@ -114,6 +114,8 @@ public class DruidDataSource extends DruidAbstractDataSource
     //
     private DruidConnectionHolder[] evictConnections;
     private DruidConnectionHolder[] keepAliveConnections;
+    // for cleaning the old references of the retained connections that have been moved forward to the new positions.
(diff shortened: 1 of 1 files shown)


zrlw:
1. Is Java's System.arraycopy() efficient for small arrays?
https://stackoverflow.com/questions/8526907/is-javas-system-arraycopy-efficient-for-small-arrays
```it will most probably be calling a JIT-specific implementation (once the JIT tags System.arraycopy as being "hot") that is not executed through the JNI interface, so it doesn't incur the normal overhead of native methods.```
2. System.arrayCopy is a hotspot jvm intrinsic function. 
https://chriswhocodes.com/hotspot_intrinsics_openjdk8.html
3. Full list of intrinsic functions in HotSpot in JDK 7, 8, 9 and 10
https://sudonull.com/post/61026-Full-list-of-intrinsic-functions-in-HotSpot-in-JDK-7-8-9-and-10
An intrinsic or intrinsic function is a function that the JIT compiler can embed instead of calling Java and JNI code for optimization purposes.

Issue #4673, opened 2022-01-28
Auto-Configuration should expose explicit datasource type, not interface


snicoll:
The following auto-configuration exposes a bean of type `javax.sql.DataSource` and provide an explicit initialization method on it:

https://github.com/alibaba/druid/blob/f937ecd43363b2df61afdc120858097cb2510580/druid-spring-boot-starter/src/main/java/com/alibaba/druid/spring/boot/autoconfigure/DruidDataSourceAutoConfigure.java#L53-L58

`javax.sql.DataSource` does not have an method named `init` so introspecting the bean definition at build time is failing. Please expose the actual type rather than the interface.

linghengqian:
@wenshao Is it possible to deal with it? This affects additional applications.

zrlw:
it might be better that removing initMethod from ```@Bean``` and calling init() at the end of DruidDataSourceWrapper#afterPropertiesSet

zrlw:
refer to org.springframework.beans.factory.BeanFactory javadocs:
```
The full set of initialization methods and their standard order is: 
1.BeanNameAware's setBeanName 
2.BeanClassLoaderAware's setBeanClassLoader 
3.BeanFactoryAware's setBeanFactory 
4.EnvironmentAware's setEnvironment 
5.EmbeddedValueResolverAware's setEmbeddedValueResolver 
6.ResourceLoaderAware's setResourceLoader(only applicable when running in an application context) 
7.ApplicationEventPublisherAware's setApplicationEventPublisher(only applicable when running in an application context) 
8.MessageSourceAware's setMessageSource(only applicable when running in an application context) 
9.ApplicationContextAware's setApplicationContext(only applicable when running in an application context) 
10.ServletContextAware's setServletContext(only applicable when running in a web application context) 
11.postProcessBeforeInitialization methods of BeanPostProcessors 
12.InitializingBean's afterPropertiesSet 
13.a custom init-method definition 
14.postProcessAfterInitialization methods of BeanPostProcessors 
```

PR #5603, opened 2023-12-14
initialize data source at DruidDataSourceWrapper#afterPropertiesSet

zrlw:
see #4673 
1. call DruidDataSource#init at the end of DruidDataSourceWrapper#afterPropertiesSet;
2. call DruidDataSource#close at DruidDataSourceWrapper#destroy.

A custom init method is an alternative to implementing InitializingBean, but DruidDataSourceWrapper is already a InitializingBean, so the initialization of  DruidDataSource could be done at DruidDataSourceWrapper#afterPropertiesSet by calling init method directly instead of setting initMethod attribute of ```@Bean``` annotation on it's  auto configuration method.


shenjianeng:
销毁的时候也调下 close 呗

zrlw:
> 销毁的时候也调下 close 呗

已加

PR #5629, opened 2023-12-18
fix: parseStatement not check SQL end

qxo:
此PR是应邀重新合并, 源PR为: #3969  

where拼写错误导致where之后条件被忽略！

之前PR #1577 虽然被合并到了主分支，但实际上执行点从来没生效过:(
而且本PR也只是处理parseStatement：加了后置断言，其它parseStatementList这一组依然存在此问题。

PR #5634, opened 2023-12-19
fix SQL statement parser support windows style SQL, check end token by default, add MySQL ELSEIF token, add SQLAlterProcedureStatement visitor

zrlw:
1. check end token at SQLStatementParser#parseStatement by default. refers to #5629 
2. replace ```\r\n``` with ```\n``` at SQLParserUtils#createSQLStatementParser to support windows style newline SQL.
3. bypass HINT token before parsing COMMA token.
4. add MySQL ELSEIF token and fix it's output format.
5. add SQLAlterProcedureStatement visitor.
6. modify WallProvider to support HintAllow option.
7. fix unit tests and test resources.


lizongbo:
我先合并了pr，一般情况下解析sql再拼成sql去执行还对字符串字段值中的\r\n敏感的可能性较低，将来还是需要考虑一下这种场景的识别适配。

PR #5714, opened 2024-01-30
Fix nested table type creation parsing in declaration

ZhengguanLi:
```sql
DECLARE
	TYPE Foursome IS TABLE OF VARCHAR2(15);
BEGIN
       ...
END;
```

Is translated to

```sql
DECLARE
	VARCHAR2 VARCHAR2(15);
BEGIN
        ...
END;
```

The reason is part of because the `name` variable is reassigned

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
@@ -1985,7 +1985,7 @@ public class OracleStatementParser extends SQLStatementParser {
                     lexer.nextToken();
                     accept(Token.OF);
 
-                    name = this.exprParser.name();
+                    SQLName sqlName = this.exprParser.name();
 
                     if (lexer.token() == Token.PERCENT) {
                         lexer.nextToken();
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
```diff
@@ -1985,24 +1985,24 @@ private void parserParameters(List<SQLParameter> parameters, SQLObject parent) {
                     lexer.nextToken();
                     accept(Token.OF);
 
-                    name = this.exprParser.name();
+                    SQLName sqlName = this.exprParser.name();
```
ZhengguanLi, line 1989: Take `TYPE Foursome IS TABLE OF VARCHAR2(15);` as example, the name should be the declared variable name `Foursome`, but since the the override, it became the `VARCHAR2`

core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
```diff
@@ -1985,24 +1985,24 @@ private void parserParameters(List<SQLParameter> parameters, SQLObject parent) {
...
                             acceptIdentifier("TYPE");
-                            typeName = "TABLE OF " + name.toString() + "%TYPE";
+                            typeName = "TABLE OF " + sqlName.toString() + "%TYPE";
                         }
 
                         dataType = new SQLDataTypeImpl(typeName);
                     } else if (lexer.token() == Token.LPAREN) {
                         lexer.nextToken();
-                        String typeName = name.toString();
+                        String typeName = "TABLE OF " + sqlName.toString();
```
ZhengguanLi, line 2005: Add `TABLE OF` to indicate it is a nested table. Not sure if we have a better way to take care of this, Druid only has `tableOf` in class `OracleCreateTypeStatement.java`

lizongbo:
CI Test一直失败，麻烦核实一下

ZhengguanLi:
好像原来的test cases就是有问题的 已经update了... 还有请问是否可以在code中使用JDK17的新特性 比如text block

PR #5725, opened 2024-02-04
Fix Oracle SELECT FOR UPDATE of output and PG SET issue

ZhengguanLi:
Changes:
- PG: If there is no "FROM" clause, avoid adding `DUAL` table as PG does not have it
- Fix Oracle, PG `SELECT FOR UPDATE OF` output:
  Should be `SELECT ... FOR UPDATE OF col1, col2;` instead of  `SELECT ... FOR UPDATE OF(col1, col2);`
  - Update some test cases using `FOR UPDATE OF(col...)`
- Restructure `visit(SQLSetStatement x)` for PG
- For `SQLSetStatement` output in PG, use `SET` only if the original query is using `SET`, related to issue #5663. It avoid s making P/SQL declaration to become SET. For example, `DECLARE a := 1 ...` to `DECLARE SET a := 1`.
  - Fix some test cases

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/postgresql/visitor/PGOutputVisitor.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/postgresql/visitor/PGOutputVisitor.java
@@ -628,7 +628,9 @@ public class PGOutputVisitor extends SQLASTOutputVisitor implements PGASTVisitor
 
     @Override
     public boolean visit(SQLSetStatement x) {
-        print0(ucase ? "SET " : "set ");
+        if (x.isUseSet()) {
+            print0(ucase ? "SET " : "set ");
+        }
 
         SQLSetStatement.Option option = x.getOption();
         if (option != null) {
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/postgresql/visitor/PGOutputVisitor.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/postgresql/visitor/PGOutputVisitor.java
@@ -645,13 +645,20 @@ public class PGOutputVisitor extends SQLASTOutputVisitor implements PGASTVisitor
 
     @Override
     public boolean visit(SQLAssignItem x) {
+        if (!(x.getParent() instanceof SQLSetStatement)) {
+            return super.visit(x);
+        }
+
         x.getTarget().accept(this);
         SQLExpr value = x.getValue();
         if (x.getTarget() instanceof SQLIdentifierExpr
             && ((SQLIdentifierExpr) x.getTarget()).getName().equalsIgnoreCase("TIME ZONE")) {
             print(' ');
         } else {
-            if (x.getParent() instanceof SQLSetStatement && !((SQLSetStatement) x.getParent()).isUseSet()) {
+            if (!((SQLSetStatement) x.getParent()).isUseSet()) {
+                print0(" := ");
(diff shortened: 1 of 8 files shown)


ZhengguanLi:
```
"SELECT rbi.rma_id, rbi.last_update_time\n" +
                        "FROM ktv_epm.ktv_eq_rprboarditem rbi\n" +
                        "JOIN ktv_epm.ktv_eq_rma_header rma ON rma.rma_id = rbi.rma_id \n" +
                        "\tJOIN ktv_epm.ktv_eq_rprsenditem rsi ON rsi.e_rprboardid = rbi.e_rprboardid\n" +
                        "AND rsi.e_boardid = rbi.e_boardid \n" +
                        "WHERE rbi.rma_id > ?\n" +
                        "\tAND rsi.e_rprsendid = ?\n" +
                        "\tAND rsi.e_boardid IN (?)\n" +
                        "FOR UPDATE rbi.rma_id\n";
```

Druid wrongly parses this kind of query to 2 statements, the last one is `rbi.rma_id`, the rest is the 1st one, that's why some tests are failing...

core/src/test/java/com/alibaba/druid/bvt/sql/oracle/create/OracleCreateViewTest13.java
```diff
@@ -55,7 +55,7 @@ public void test_types() throws Exception {
                         "WHERE rbi.rma_id > ?\n" +
                         "\tAND rsi.e_rprsendid = ?\n" +
                         "\tAND rsi.e_boardid IN (?)\n" +
-                        "FOR UPDATE(rbi.rma_id)",//
+                        "FOR UPDATE OF rbi.rma_id",//
```
ZhengguanLi, line 58: I don't find any place of Oracle doc saying it support `FOR UPDATE(col)` syntax, correct if I am wrong

core/src/test/java/com/alibaba/druid/bvt/sql/oracle/create/OracleCreateTriggerTest6.java
```diff
@@ -120,8 +120,8 @@ public void test_0() throws Exception {
                         "\tvar_nRecordCount number;\n" +
                         "\tvar_sState varchar2(16);\n" +
                         "BEGIN\n" +
-                        "\tSET var_sState := :NEW.State;\n" +
-                        "\tSET var_dFirstCheckTime := :NEW.Commiteddate;\n" +
+                        "\tvar_sState := :NEW.State;\n" +
+                        "\tvar_dFirstCheckTime := :NEW.Commiteddate;\n" +
```
ZhengguanLi, line 124: Also I don't see where SET command can be used in Oracle declaration...

core/src/main/java/com/alibaba/druid/sql/dialect/postgresql/visitor/PGOutputVisitor.java
```diff
@@ -628,44 +628,48 @@ public boolean visit(PGAlterSchemaStatement x) {
...
             print(option.name());
             print(' ');
         }
 
-        List<SQLAssignItem> items = x.getItems();
-        for (int i = 0; i < items.size(); i++) {
-            if (i != 0) {
-                print0(", ");
-            }
+        printAndAccept(x.getItems(), ", ");
```
ZhengguanLi, line 641: This method from parent class can do the for loop for us, reuse it

PR #5734, opened 2024-02-15
Support Oracle REVERSE FOR Loop

ZhengguanLi:
Support Oracle REVERSE FOR Loop

Changes:
- Add `REVERSE` token in `core/src/main/java/com/alibaba/druid/sql/parser/Token.java`
- Add `REVERSE` token in `core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleLexer.java`
- Add `isReverse` field in `core/src/main/java/com/alibaba/druid/sql/ast/statement/SQLForStatement.java`
- Support `REVERSE` FOR Loop parsing and output
- Add test `core/src/test/java/com/alibaba/druid/bvt/sql/oracle/block/OracleBlockTest4.java`

Other Changes:
- Fix some warnings in `core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleLexer.java` and `core/src/main/java/com/alibaba/druid/sql/parser/Keywords.java`

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/ast/statement/SQLForStatement.java
+++ b/core/src/main/java/com/alibaba/druid/sql/ast/statement/SQLForStatement.java
@@ -11,6 +11,7 @@ import java.util.List;
 
 public class SQLForStatement extends SQLStatementImpl {
     protected SQLName index;
+    protected boolean isReverse;
     protected SQLExpr range;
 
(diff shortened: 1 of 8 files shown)


ZhengguanLi:
Looks like `REVERSE` can be seen both as token in FOR LOOP and as Identifier(`REVERSE('abc')`) in function like from MySQL as an example in `EvalMethodReverseTest`. Need a way to separate them in the code to resolve it...

PR #6170, opened 2024-10-14
Support ignore/respect nulls in bigQuery.

lingo-xp:


core/src/main/java/com/alibaba/druid/sql/ast/expr/SQLAggregateExpr.java
```diff
@@ -37,20 +37,27 @@ public class SQLAggregateExpr extends SQLMethodInvokeExpr implements Serializabl
     protected SQLOrderBy orderBy;
     protected SQLExpr limit;
     protected boolean withinGroup;
-    protected Boolean ignoreNulls;
+    protected boolean ignoreNulls;
```
wenshao, line 40: 补上注释for bigquery

PR #6185, opened 2024-10-21
Remove useless code.

lingo-xp:
…all when setting null.


lingo-xp:
Remove useless code in SQLDeleteStatement to avoid ambiguous method call when setting null. 
```
table.setTableSource(null);
```

PR #6243, opened 2024-11-22
Support RETURNING schema statistics in INSERT statement

ZhengguanLi:
Change:
- Support `RETURNING` schema statistics in `INSERT` statement(PLSQL)
- Support `FROM` clause schema statistics in method(`TRIM` specifically)


wenshao:
需要修复下代码冲突

PR #6335, opened 2025-01-22
fix insert parse \' in varchar value.

dingxiaobo:
fix bad case:

```sql
INSERT INTO tb (id, c0) VALUES(1, 'CREATE TABLE tb (
           id bigint(20) NOT NULL AUTO_INCREMENT,
            c0 varchar(40) NOT NULL COMMENT \'c0\',
            PRIMARY KEY (id)
            )');
```

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/parser/Lexer.java
+++ b/core/src/main/java/com/alibaba/druid/sql/parser/Lexer.java
@@ -514,7 +514,11 @@ public class Lexer {
         this.startPos = pos;
         if (ch == '\'') {
             bufPos = 0;
-            scanString();
+            if (dbType == DbType.mysql) {
+                scanString2();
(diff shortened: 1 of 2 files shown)


dingxiaobo:
rollback 6d8dd333a3a63e6e2615c9376e023086d0355263

dingxiaobo:
Use `SQLUtils.parseStatements` instead, problem solved.

```
SQLUtils.parseStatements("INSERT INTO tb (id, c0) VALUES(1, 'CREATE TABLE tb (\n"
            + "id bigint(20) NOT NULL AUTO_INCREMENT,\n"
            + " c0 varchar(40) NOT NULL COMMENT \\'c0\\',\n"
            + " PRIMARY KEY (id)\n"
            + " )');", DbType.mysql)

```

PR #6343, opened 2025-02-08
support db2/sqlserver some ddl sql

zycgit:
db2
rename table xxx
create table xxx as xxx with data;
alter table user_table drop address;
alter table abc drop unique xxx;
alter table abc drop foreign key xxx;

sqlserver


suger-no:
sqlserver
create schema
create schema create table xxxxxx
create schema create view xxxxx
create schema authorization
drop schema

core/src/main/java/com/alibaba/druid/sql/dialect/db2/ast/stmt/DB2CreateTableStatement.java
```diff
@@ -25,10 +26,27 @@ public class DB2CreateTableStatement extends SQLCreateTableStatement implements
     private boolean dataCaptureNone;
     private boolean dataCaptureChanges;
 
+    private AsSelectWith selectWith;
```
wenshao, line 29: 这种为什么不用一个boolean解决？

zycgit: 因为还有其它方式，本次只支持了其中一种。如：
create table new_table_name1 as (select * from table1) definition only;
create table new_table_name2 as (select * from table1) with data;
create table new_table_name3 as (select * from table1) WITH no data;

PR #6417, opened 2025-05-07
Fix solve the issue in SQL where the < operator is mistakenly recogni…

shallowShark123:
…zed as <@ when used jointly with user-defined variables.
In SQL, <@ is a special binary operator. However, in MySQL parsing, if a space is added between < and @, MySQL will report a parsing error when executing. This means that '<@' does not need to be identified separately in MySQL as LT_MONKEYS_AT type.
In addition, when the < operator is mixed with user-defined variables, the token type identified by the current lexical parsing is incorrect.
For example:
select * from aaa where id > @sid
The correct identification method is that @sid is a user-defined variable, and the id field is compared with the variable using the binary operator >.
Current identification method:
>@ is identified as the binary operator Array_ContainedBy, which is incorrect.

### DIFF
--- a/core/src/test/java/com/alibaba/druid/mysql/MysqlVarantRefTest.java
+++ b/core/src/test/java/com/alibaba/druid/mysql/MysqlVarantRefTest.java
@@ -3,16 +3,23 @@ package com.alibaba.druid.mysql;
 import com.alibaba.druid.DbType;
 import com.alibaba.druid.sql.SQLUtils;
 import com.alibaba.druid.sql.ast.SQLStatement;
+import com.alibaba.druid.sql.ast.expr.SQLBinaryOpExpr;
+import com.alibaba.druid.sql.ast.expr.SQLBinaryOperator;
 import com.alibaba.druid.sql.ast.expr.SQLVariantRefExpr;
 import com.alibaba.druid.sql.ast.statement.SQLSetStatement;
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/alibaba/druid/mysql/MysqlVarantRefTest.java
```diff
@@ -150,4 +157,54 @@ public void test6() {
...
+        SQLStatementParser parser = new MySqlStatementParser(sql);
+        List<SQLStatement> stmtList = parser.parseStatementList();
+
+        SQLSetStatement result0 = (SQLSetStatement) stmtList.get(0);
+        SQLVariantRefExpr resultExpr0 = (SQLVariantRefExpr) result0.getItems().get(0).getTarget();
+        Assert.assertEquals("@sid", resultExpr0.getName());
+
+        AtomicReference<SQLVariantRefExpr> resultExpr1Ref = new AtomicReference<>();
+        AtomicInteger variableCnt = new AtomicInteger(0);
+        AtomicInteger arrayContainerByOperatorCnt = new AtomicInteger(0);
```
mengnankkkk, line 173: 可补充更多边界场景，如带空格、注释、复杂表达式等，提升健壮性。
建议补充负面测试，如非法 SQL、混合操作符等。

PR #6460, opened 2025-06-19
Add get multi level parent method.

lingo-xp:


core/src/main/java/com/alibaba/druid/sql/ast/SQLObjectImpl.java
```diff
@@ -108,6 +108,16 @@ public String toString() {
     public SQLObject getParent() {
         return parent;
     }
+    public SQLObject getParent(int level) {
+        SQLObject parent = this;
```
wenshao, line 112: 检查下输入参数level > 0

PR #6628, opened 2026-03-25
fix: MapComparator overflow causes SQL stat sort to fail

### LINKED ISSUE #6624: [BUG] sql语句监控页面，按条件排序，倒序无效
### Database Type

MySQL

### Database Version

MySQL5.7

### Druid Version

1.2.28

### JDK Version

OpenJDK17

### Error SQL

无论怎么选择都是升序，无法降序，图标在切换，数据没变化
springboot3.5
Image

Image

### Testcase Code

_No response_

### Stacktrace Info

_No response_

### Error Info

_No response_

daguimu:
## Problem

SQL monitoring page sort doesn't work — clicking column headers toggles the ▲▼ icons but the data order doesn't change. Descending sort has no effect.

**Root cause:** `MapComparator` uses `(int)(longA - longB)` for Long/Date/Number comparisons. When values are large (which is common for cumulative SQL stats like `ExecuteCount` and `TotalTime`), the subtraction overflows `int` range, producing wrong comparison results that violate the `Comparator` contract.

Example: `(int)(3000000000L - 1L)` = `-1294967297` (negative! should be positive)

This causes `Collections.sort` to produce unpredictable ordering that doesn't change between ascending and descending mode — exactly matching the reported behavior.

## Fix

Replace all 3 overflow-prone comparisons with safe alternatives:

| Before (overflow) | After (safe) |
|---|---|
| `(int)(long1 - long2)` | `Long.compare(long1, long2)` |
| `(int)(double1 - double2)` | `Double.compare(double1, double2)` |
| `(int)(date1.getTime() - date2.getTime())` | `Long.compare(date1.getTime(), date2.getTime())` |

The `Double.compare` fix also resolves a precision bug where small differences (e.g., `100.5 - 100.0 = 0.5`) were truncated to `0` by the `(int)` cast, treating close values as equal.

## Tests Added

7 new test cases in `MapComparatorOverflowTest.java`:
- `test_long_sort_ascending_large_values` — ASC sort with Long.MAX_VALUE range
- `test_long_sort_descending_large_values` — DESC sort with Long.MAX_VALUE range
- `test_long_sort_desc_differs_from_asc` — verifies DESC produces reverse of ASC
- `test_long_overflow_values` — specific overflow case: 3B vs 1
- `test_date_sort_large_time_difference` — Date comparison with epoch vs now
- `test_number_small_difference_precision` — 100.5 vs 100.0 precision
- `test_random_long_sort_consistency` — 100 random Longs, no TimSort contract violation

All 4 existing `MapComparatorTest` tests continue to pass.

## Impact

Fixes SQL statement monitoring page sort for all stat columns (`ExecuteCount`, `TotalTime`, `MaxTimespan`, etc.). The fix is minimal (3 line changes) with no behavioral change for values within `int` range.

Fixes #6624

### DIFF
--- a/core/src/main/java/com/alibaba/druid/util/MapComparator.java
+++ b/core/src/main/java/com/alibaba/druid/util/MapComparator.java
@@ -34,7 +34,7 @@ public class MapComparator<K extends Object, V extends Object> implements Compar
     }
 
     private int compare(Number o1, Number o2) {
-        return (int) (o1.doubleValue() - o2.doubleValue());
+        return Double.compare(o1.doubleValue(), o2.doubleValue());
     }
 
     private int compare(String o1, String o2) {
@@ -42,7 +42,7 @@ public class MapComparator<K extends Object, V extends Object> implements Compar
     }
 
     private int compare(Date o1, Date o2) {
-        return (int) (o1.getTime() - o2.getTime());
+        return Long.compare(o1.getTime(), o2.getTime());
     }
 
     @Override
@@ -99,7 +99,7 @@ public class MapComparator<K extends Object, V extends Object> implements Compar
         }
 
         if (v1 instanceof Long) {
-            return (int) (((Long) v1).longValue() - ((Number) v2).longValue());
+            return Long.compare(((Long) v1).longValue(), ((Number) v2).longValue());
         }
 
         if (v1 instanceof Number) {
--- a/core/src/test/java/com/alibaba/druid/util/MapComparatorOverflowTest.java
+++ b/core/src/test/java/com/alibaba/druid/util/MapComparatorOverflowTest.java
@@ -1,0 +1,151 @@
+package com.alibaba.druid.util;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.*;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Fix MapComparator sorting failures due to integer overflow in Long/Date comparisons.
+ * <p>
+ * The previous implementation used {@code (int)(longA - longB)} which overflows for large
+ * values, violating the Comparator contract and causing incorrect sort order.
+ * This manifests as the SQL monitoring page sort not working (descending has no effect).
+ *
+ * @see <a href="https://github.com/alibaba/druid/issues/6624">Issue #6624</a>
+ */
+public class MapComparatorOverflowTest {
+    @Test
+    public void test_long_sort_ascending_large_values() {
+        List<Map<String, Object>> list = new ArrayList<>();
+        long[] values = {Long.MAX_VALUE, 1L, Long.MAX_VALUE / 2, 100L, Long.MAX_VALUE - 1};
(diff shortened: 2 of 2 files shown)


core/src/test/java/com/alibaba/druid/util/MapComparatorOverflowTest.java
```diff
@@ -0,0 +1,151 @@
...
+ * <p>
+ * The previous implementation used {@code (int)(longA - longB)} which overflows for large
+ * values, violating the Comparator contract and causing incorrect sort order.
+ * This manifests as the SQL monitoring page sort not working (descending has no effect).
+ *
+ * @see <a href="https://github.com/alibaba/druid/issues/6624">Issue #6624</a>
+ */
+public class MapComparatorOverflowTest {
+    @Test
+    public void test_long_sort_ascending_large_values() {
```
wenshao, line 20: **[Suggestion]** Test class is unreachable by surefire — the include pattern in `core/pom.xml` is `**/bvt/**/*.java`, but this test is in package `com.alibaba.druid.util`, which does not match. Running `mvn test -pl core` or the CI pipeline will compile the test but never execute it. — Concrete cost: if the `Long.compare`/`Double.compare` fix is later reverted or regressed, no CI run will catch it — the 7 tests written to detect exactly that regression will silently not run.

```suggestion
// Move this file to:
// core/src/test/java/com/alibaba/druid/bvt/util/MapComparatorOverflowTest.java
// and change the package declaration to:
package com.alibaba.druid.bvt.util;
```

core/src/main/java/com/alibaba/druid/util/MapComparator.java
```diff
@@ -99,7 +99,7 @@ private int compare_0(Map<K, V> o1, Map<K, V> o2) {
         }
 
         if (v1 instanceof Long) {
-            return (int) (((Long) v1).longValue() - ((Number) v2).longValue());
+            return Long.compare(((Long) v1).longValue(), ((Number) v2).longValue());
```
wenshao, line 102: **[Suggestion]** The interaction between the fixed Long branch and the Number branch introduces a Comparator antisymmetry violation for mixed Long/non-Long Number values. `compare(Long(5), Double(5.9))` returns 0 (Long branch truncates v2 via `.longValue()`), but `compare(Double(5.9), Long(5))` returns 1 (Number branch uses precise `Double.compare`). This violates `sgn(compare(x,y)) == -sgn(compare(y,x))`. — Failure scenario: if a sorted list contains maps where the same key holds a `Long` in some rows and a `Double`/`Float` in others, TimSort can throw `IllegalArgumentException: Comparison method violates its general contract!`.

```suggestion
                if (v1 instanceof Long && v2 instanceof Long) {
                    return Long.compare((Long) v1, (Long) v2);
```

This ensures mixed-type pairs fall through to the symmetric `Double.compare` path in both directions.

wenshao (commented):

PR #6629, opened 2026-03-25
fix: Oracle PL/SQL TYPE ... IS RECORD parsing in package body

### LINKED ISSUE #6589: Oracle脚本语句解析报错
使用了druid 1.2.27版本，调用SQLUtils.parseStatements(content, dbType);解析sql语句报错：
syntax error, expect REF, actual TABLE, pos 1552, line 69, column 30, token TABLE

[44851-1.txt](https://github.com/user-attachments/files/25043425/44851-1.txt)

实际上这是个正确的oracle 脚本语句

daguimu:
## Problem

Oracle PL/SQL parser throws `ParserException: TODO` when encountering `TYPE ... IS RECORD(...)` declarations inside package bodies, procedures, or DECLARE blocks.

```sql
-- This fails with: TODO : ... token IDENTIFIER record
CREATE OR REPLACE PACKAGE BODY pkg_test IS
    PROCEDURE test_proc AS
        TYPE T_Rec IS RECORD(
            id NUMBER,
            name VARCHAR2(100));
    BEGIN
        NULL;
    END test_proc;
END pkg_test;
```

**Root cause:** In `OracleStatementParser` (line ~2129), the `TYPE ... IS` handler only supported `REF CURSOR`, `TABLE OF`, and `VARRAY` — `RECORD` was missing, falling through to a `throw new ParserException("TODO")`.

## Fix

Add `RECORD` type parsing to the PL/SQL declaration section, following the same implementation pattern already used at line ~2940 in the same file. The fix parses `RECORD(field1 type1, field2 type2, ...)` using `SQLRecordDataType` and `parseColumn()` which already exist in the codebase.

## Tests Added

6 test cases in `Issue6589.java`:
- `test_type_is_record_in_package_body_procedure` — RECORD type inside a procedure in a package body
- `test_type_is_record_in_declare_block` — RECORD type in a standalone DECLARE block
- `test_type_is_record_multiple_fields` — RECORD with 4 fields of different types
- `test_ref_cursor_still_works` — regression: REF CURSOR still works
- `test_table_of_still_works` — regression: TABLE OF still works
- `test_package_spec_and_body_with_record` — package spec + body with `/` separator

All 15 existing Oracle issue tests pass.

## Note

The original issue's full SQL file also uses `BULK COLLECT INTO` and other advanced PL/SQL features that remain unsupported. This PR addresses the `TYPE IS RECORD` parsing specifically.

Fixes #6589

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
@@ -2126,6 +2126,25 @@ public class OracleStatementParser extends SQLStatementParser {
                     } else {
                         throw new ParserException("TODO : " + lexer.info());
                     }
+                } else if (lexer.identifierEquals(FnvHash.Constants.RECORD)) {
+                    lexer.nextToken();
+
+                    SQLRecordDataType recordDataType = new SQLRecordDataType();
+
+                    accept(Token.LPAREN);
+                    for (; ; ) {
+                        SQLColumnDefinition column = this.exprParser.parseColumn();
+                        recordDataType.addColumn(column);
+                        if (lexer.token() == Token.COMMA) {
+                            lexer.nextToken();
+                            continue;
+                        }
+                        break;
+                    }
+                    accept(Token.RPAREN);
+
+                    dataType = recordDataType;
+                    dataType.setDbType(dbType);
                 } else {
                     throw new ParserException("TODO : " + lexer.info());
                 }
--- a/core/src/test/java/com/alibaba/druid/bvt/sql/oracle/issues/Issue6589.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/sql/oracle/issues/Issue6589.java
@@ -9,39 +9,115 @@ import org.junit.jupiter.api.Test;
 import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
- * @see <a href="https://github.com/alibaba/druid/issues/6589">Issue 6589</a>
- * Oracle PL/SQL parsing fails on TYPE ... IS TABLE OF inside package body.
+ * Fix Oracle PL/SQL TYPE ... IS RECORD(...) parsing in package body and declare blocks.
+ * <p>
+ * The parser previously threw "TODO" when encountering RECORD after TYPE ... IS,
+ * only supporting REF CURSOR, TABLE OF, and VARRAY.
+ *
+ * @see <a href="https://github.com/alibaba/druid/issues/6589">Issue #6589</a>
  */
 public class Issue6589 {
     @Test
-    public void test_type_is_table_of_in_package_body() {
-        String sql = "CREATE OR REPLACE PACKAGE BODY my_pkg AS\n"
-                + "  type T_Validator_List is table of varchar2(200);\n"
-                + "  PROCEDURE my_proc IS\n"
-                + "  BEGIN\n"
-                + "    NULL;\n"
-                + "  END;\n"
-                + "END;";
+    public void test_type_is_record_in_package_body_procedure() {
+        String sql = "create or replace PACKAGE BODY pkg_test IS\n"
+                + "    procedure test_proc as\n"
+                + "        type T_Rec is record(\n"
+                + "            DeferredDate date,\n"
+                + "            LicenseText  varchar2(300));\n"
+                + "        type T_Rec_List is table of T_Rec;\n"
+                + "        v_items T_Rec_List;\n"
+                + "    begin\n"
+                + "        null;\n"
+                + "    end test_proc;\n"
+                + "END pkg_test;";
 
         SQLStatementParser parser = SQLParserUtils.createSQLStatementParser(sql, DbType.oracle);
-        List<SQLStatement> statementList = parser.parseStatementList();
-        assertEquals(1, statementList.size());
+        List<SQLStatement> stmtList = parser.parseStatementList();
+        assertEquals(1, stmtList.size());
+        assertNotNull(stmtList.get(0));
     }
 
     @Test
-    public void test_type_is_table_of_index_by_in_package_body() {
-        String sql = "CREATE OR REPLACE PACKAGE BODY my_pkg AS\n"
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
```diff
@@ -2126,6 +2126,25 @@ protected void parserParameters(List<SQLParameter> parameters, SQLObject parent)
                     } else {
                         throw new ParserException("TODO : " + lexer.info());
                     }
+                } else if (lexer.identifierEquals(FnvHash.Constants.RECORD)) {
+                    lexer.nextToken();
+
+                    SQLRecordDataType recordDataType = new SQLRecordDataType();
```
wenshao, line 2132: **[Critical]** The new branch creates a `SQLRecordDataType` without setting its `name` (inherited from `SQLDataTypeImpl`, defaults to `null`) and stores it on a `SQLParameter` via `parameter.setDataType(...)`. When the AST is serialized, `SQLASTOutputVisitor.visit(SQLParameter)` calls `x.getDataType().getName().equalsIgnoreCase("CURSOR")`, which throws `NullPointerException` because `getName()` returns null. — Failure scenario: parse any `TYPE ... IS RECORD(...)` inside a package body procedure (e.g. the SQL in `test_type_is_record_in_package_body_procedure`), then call `SQLUtils.toSQLString(stmt, DbType.oracle)` or `stmt.toString()` → NullPointerException. The pre-existing RECORD path (~line 2962) stores the type on a `SQLDeclareItem`, whose visitor has an `instanceof SQLRecordDataType` guard before touching `getName()`; this new `SQLParameter` path has none, so the NPE is newly reachable by this diff. The new tests only assert statement count and never serialize the AST, so they do not catch it.

Suggested fix: add an `instanceof SQLRecordDataType` early-return in `SQLASTOutputVisitor.visit(SQLParameter)` (mirroring the existing `OracleFunctionDataType` / `OracleProcedureDataType` handling and the guard already present in `visit(SQLDeclareItem)`), and add a round-trip assertion (`SQLUtils.toSQLString`) to a test so this path is exercised.

core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleStatementParser.java
```diff
@@ -2126,6 +2126,25 @@ protected void parserParameters(List<SQLParameter> parameters, SQLObject parent)
...
+                        recordDataType.addColumn(column);
+                        if (lexer.token() == Token.COMMA) {
+                            lexer.nextToken();
+                            continue;
+                        }
+                        break;
+                    }
+                    accept(Token.RPAREN);
+
+                    dataType = recordDataType;
```
wenshao, line 2146: **[Suggestion]** This RECORD column-parsing loop is a verbatim duplicate of the existing loop at ~lines 2964–2974 (the `parseDeclare` variable-declaration path) in the same file. — Concrete cost: a future fix to RECORD field parsing (e.g. `NOT NULL` on fields, a trailing comma, or an empty `record()`) must be applied to two identical blocks; missing one makes RECORD declarations in package procedure bodies parse differently from those in standalone DECLARE blocks.

Suggested fix: extract a private helper, e.g. `private SQLRecordDataType parseRecordDataType() { ... }`, and call it from both sites.

core/src/test/java/com/alibaba/druid/bvt/sql/oracle/issues/Issue6589.java
```diff
@@ -9,39 +9,115 @@
...
+                + "    procedure get_data(p_result out t_cursor);\n"
+                + "END pkg_test;";
+
+        SQLStatementParser parser = SQLParserUtils.createSQLStatementParser(sql, DbType.oracle);
+        List<SQLStatement> stmtList = parser.parseStatementList();
+        assertEquals(1, stmtList.size());
+    }
+
+    @Test
+    public void test_table_of_still_works() {
```
wenshao, line 91: **[Suggestion]** The deleted `test_type_is_table_of_index_by_in_package_body` was the only test exercising the `INDEX BY` sub-branch of `parserParameters` (the `if (lexer.token() == Token.INDEX)` block at ~line 2090 — the same method this PR modifies). The replacement `test_table_of_still_works` covers only plain `TABLE OF`, and the surviving `INDEX BY` tests elsewhere go through different code paths (`parseDeclare` / `CREATE TYPE`). — Concrete cost: a future change to the `parserParameters` `INDEX BY` branch would ship without any direct regression test.

Suggested fix: add a test restoring the `INDEX BY BINARY_INTEGER` variant in a package body, e.g.:

```java
@Test
public void test_table_of_index_by_still_works() {
    String sql = "CREATE OR REPLACE PACKAGE BODY my_pkg AS\n"
            + "  type T_Name_List is table of varchar2(100) index by binary_integer;\n"
            + "END;";
    SQLStatementParser parser = SQLParserUtils.createSQLStatementParser(sql, DbType.oracle);
    assertEquals(1, parser.parseStatementList().size());
}
```

core/src/test/java/com/alibaba/druid/bvt/sql/oracle/issues/Issue6589.java
```diff
@@ -9,39 +9,115 @@
...
+                + "        null;\n"
+                + "    end test_proc;\n"
+                + "END pkg_test;";
 
         SQLStatementParser parser = SQLParserUtils.createSQLStatementParser(sql, DbType.oracle);
-        List<SQLStatement> statementList = parser.parseStatementList();
-        assertEquals(1, statementList.size());
+        List<SQLStatement> stmtList = parser.parseStatementList();
+        assertEquals(1, stmtList.size());
+        assertNotNull(stmtList.get(0));
```
wenshao, line 41: **[Suggestion]** All six tests assert only that parsing returns a statement (`size()` / non-null); none inspects the parsed AST, so the core behaviour — columns being added to `SQLRecordDataType` — is unguarded. — Concrete cost: deleting `recordDataType.addColumn(column)` from the parser still passes every test (the loop advances the lexer, commas are consumed, `RPAREN` is accepted, and the statement count is unchanged), so a zero-column `SQLRecordDataType` would ship undetected.

Suggested fix: in at least `test_type_is_record_in_package_body_procedure` and `test_type_is_record_multiple_fields`, navigate to the `SQLRecordDataType` and assert column count / names, e.g. `assertEquals(2, record.getColumns().size())`.

wenshao (changes requested):

Issue #6534, opened 2025-09-19
[BUG] 不支持 member of


chengmaoning:
### Database Type

mysql

### Database Version

8.4.6

### Druid Version

1.2.23

### JDK Version

OpenJDK8

### Error SQL

`select p.id, p.create_time, p.update_time, p.name, p.code, p.biz_domain, p.proj_desc, p.owner, p.status, p.priority, p.start_time, p.end_time, p.remark, p.update_by, p.create_by, p.cancel_remark, p.cust_id, p.participants from project p WHERE p.del_flag = '0' and p.status in ( ? ) and ? member of(p.participants -> '$.*[*]') order by p.id desc LIMIT ?`

### Error Info

com.alibaba.druid.sql.parser.ParserException: not supported.pos 861, line 42, column 22, token IDENTIFIER member
	at com.alibaba.druid.sql.parser.SQLStatementParser.parseStatementList(SQLStatementParser.java:654)
	at com.alibaba.druid.sql.parser.SQLStatementParser.parseStatementList(SQLStatementParser.java:106)
	at com.alibaba.druid.sql.visitor.ParameterizedOutputVisitorUtils.parameterize(ParameterizedOutputVisitorUtils.java:164)
	at com.alibaba.druid.sql.visitor.ParameterizedOutputVisitorUtils.parameterize(ParameterizedOutputVisitorUtils.java:135)
	at com.alibaba.druid.filter.stat.StatFilter.mergeSql(StatFilter.java:158)
	at com.alibaba.druid.filter.stat.StatFilter.createSqlStat(StatFilter.java:692)
	at com.alibaba.druid.filter.stat.StatFilter.statementPrepareAfter(StatFilter.java:339)
	at com.alibaba.druid.filter.FilterEventAdapter.connection_prepareStatement(FilterEventAdapter.java:118)
	at com.alibaba.druid.filter.FilterChainImpl.connection_prepareStatement(FilterChainImpl.java:547)
	at com.alibaba.druid.proxy.jdbc.ConnectionProxyImpl.prepareStatement(ConnectionProxyImpl.java:328)
	at com.alibaba.druid.pool.DruidPooledConnection.prepareStatement(DruidPooledConnection.java:369)
	at sun.reflect.GeneratedMethodAccessor227.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

Issue #4848, opened 2022-06-27
ParserException，不支持mysql member of查询json字段


zhakumi:
** 1.环境**
数据库：8.0.22
ORM： mybatis
DRUID:1.1.12
2.sql，类似下面这种
select count(bu.id) from a  WHERE   'ddd' member of (a.extra_info->'$.outCustomerId')

mysql中支持member of关键字，druid不支持

rg.mybatis.spring.MyBatisSystemException: nested exception is org.apache.ibatis.executor.ExecutorException: Error preparing statement.  Cause: com.alibaba.druid.sql.parser.ParserException: syntax error, error in :'mber of (info.extra_info->'$.outUse', expect IDENTIFIER, actual IDENTIFIER pos 393, line 25, column 28, token IDENTIFIER of
	at org.mybatis.spring.MyBatisExceptionTranslator.translateExceptionIfPossible(MyBatisExceptionTranslator.java:92)
	at org.mybatis.spring.SqlSessionTemplate$SqlSessionInterceptor.invoke(SqlSessionTemplate.java:440)
	at com.sun.proxy.$Proxy133.selectList(Unknown Source)
	at org.mybatis.spring.SqlSessionTemplate.selectList(SqlSessionTemplate.java:223)
	at com.baomidou.mybatisplus.core.override.MybatisMapperMethod.executeForMany(MybatisMapperMethod.java:173)
	at com.baomidou.mybatisplus.core.override.MybatisMapperMethod.execute(MybatisMapperMethod.java:78)
	at com.baomidou.mybatisplus.core.override.MybatisMapperProxy$PlainMethodInvoker.invoke(MybatisMapperProxy.java:148)
	at com.baomidou.mybatisplus.core.override.MybatisMapperProxy.invoke(MybatisMapperProxy.java:89)
	at com.sun.proxy.$Proxy154.listCustomer(Unknown Source)
	at com.wakedata.wd.scrm.base.app.customer.query.CustomerInfoPageQryExe.execute(CustomerInfoPageQryExe.java:53)
	at com.wakedata.wd.scrm.base.app.customer.api.ScrmBaseCustomerRpcServiceImpl.pageCustomerInfo(ScrmBaseCustomerRpcServiceImpl.java:147)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

Issue #4358, opened 2021-07-08
druid 1.2.6 不支持 mysql 8.0  member of 


badouyuren:
java.sql.SQLException: sql injection violation, dbType mysql, , druid-version 1.2.5, syntax error: syntax error, error in :'mber of(post_json->'$.ids'), pos 43, line 1, column 45, token IDENTIFIER of : select * from beauty_similar where ? member of(post_json->'$.ids')

c-h-h:
which version will support this function?

Msimplehub:
可以先使用 Regexp 替代用一下

PR #6636, opened 2026-04-11
fix: support member of syntax in mysql parser

H4rry217:
### Problem

currently, druid does not support the mysql`MEMBER OF` operator, so this pr add support for mysql `MEMBER OF` syntax in the MySQL parser and wall path

Related issue:
- #6534
- #3734
- #4848
- #4358

Reference:
- https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html

###

### Test
Validated with:

```
mvn -pl core "-Dtest=ComparisonFunctionsAndOperatorsTest,MySqlWallTest100" test
```

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/mysql/parser/MySqlExprParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/mysql/parser/MySqlExprParser.java
@@ -561,6 +561,37 @@ public class MySqlExprParser extends SQLExprParser {
     }
 
     @Override
+    public SQLExpr relationalRest(SQLExpr expr) {
+        if (lexer.token() == Token.IDENTIFIER && lexer.hashLCase() == FnvHash.Constants.MEMBER) {
+            Lexer.SavePoint mark = lexer.mark();
+            lexer.nextToken();
+
+            if (lexer.token() == Token.OF || lexer.identifierEquals("OF")) {
+                lexer.nextToken();
+
+                SQLExpr rightExp;
+                if (lexer.token() == Token.LPAREN) {
+                    lexer.nextToken();
+                    rightExp = expr();
+                    accept(Token.RPAREN);
+                    if (rightExp instanceof SQLExprImpl) {
+                        ((SQLExprImpl) rightExp).setParenthesized(true);
+                    }
+                    rightExp = primaryRest(rightExp);
+                } else {
+                    rightExp = bitOr();
+                }
+
+                expr = new SQLBinaryOpExpr(expr, SQLBinaryOperator.MemberOf, rightExp, dbType);
+                return super.relationalRest(expr);
+            }
+            lexer.reset(mark);
+        }
+
+        return super.relationalRest(expr);
+    }
+
+    @Override
     protected SQLExpr parseSelectItemRest(String ident, long hash_lower) {
         SQLExpr expr = null;
         if (lexer.identifierEquals(FnvHash.Constants.COLLATE)
--- a/core/src/test/java/com/alibaba/druid/bvt/sql/mysql/ComparisonFunctionsAndOperatorsTest.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/sql/mysql/ComparisonFunctionsAndOperatorsTest.java
@@ -356,6 +356,18 @@ public class ComparisonFunctionsAndOperatorsTest {
         assertEquals("SELECT CAST(LEAST(3600, 9223372036854775808.0) AS SIGNED);", text);
     }
 
+    @Test
+    public void test_27() throws Exception {
+        String sql = "SELECT 'ab' MEMBER OF('[23, \"abc\", 17, \"ab\", 10]')";
+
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/alibaba/druid/sql/dialect/mysql/parser/MySqlExprParser.java
```diff
@@ -560,6 +560,37 @@ protected void parseIndexOptions(SQLIndexDefinition indexDefinition) {
...
+                lexer.nextToken();
+
+                SQLExpr rightExp;
+                if (lexer.token() == Token.LPAREN) {
+                    lexer.nextToken();
+                    rightExp = expr();
+                    accept(Token.RPAREN);
+                    if (rightExp instanceof SQLExprImpl) {
+                        ((SQLExprImpl) rightExp).setParenthesized(true);
+                    }
```
wenshao, line 579: **[Critical]** `setParenthesized(true)` is ineffective on the MySQL output path — `MySqlOutputVisitor.visit(SQLCharExpr, boolean)` (line 721) overrides the base class without checking `isParenthesized()`, so the rendered SQL drops the parentheses that MySQL's `MEMBER OF(expr)` syntax mandates. — Failure scenario: parse `SELECT 'ab' MEMBER OF('[23, ...]')` then render via `SQLUtils.toSQLString(stmts, DbType.mysql)` → output is `MEMBER OF '[23, ...]'` (no parens) → MySQL rejects with a syntax error. Any round-trip workflow (formatting, rewriting, proxying) silently produces invalid SQL. The test `test_27` asserts this incorrect output, blessing the bug.

Suggested fix: add `MemberOf`-specific handling in `SQLASTOutputVisitor.visitorBinaryRight` to always wrap the right operand in parentheses:
```java
} else if (op == SQLBinaryOperator.MemberOf) {
    print('(');
    printExpr(right, parameterized);
    print(')');
}
```
Then update the test expectation to `"SELECT 'ab' MEMBER OF ('[23, \"abc\", 17, \"ab\", 10]')"`.

core/src/test/java/com/alibaba/druid/bvt/sql/mysql/ComparisonFunctionsAndOperatorsTest.java
```diff
@@ -356,6 +356,18 @@ public void test_26() throws Exception {
         assertEquals("SELECT CAST(LEAST(3600, 9223372036854775808.0) AS SIGNED);", text);
     }
 
+    @Test
+    public void test_27() throws Exception {
+        String sql = "SELECT 'ab' MEMBER OF('[23, \"abc\", 17, \"ab\", 10]')";
```
wenshao, line 361: **[Suggestion]** Tests do not cover the JSON path expression shape reported in the motivating issues — (1) #6534, (2) #4848, (3) #4358 all use `MEMBER OF(column -> 'json_path')` with a `->` (SubGt) expression inside the parens; this test uses a plain string literal. — Concrete cost: a future refactor of `relationalRest` or the `expr()` call chain could break parsing of `MEMBER OF(expr -> 'path')` — the exact shape users reported — while this test stays green.

core/src/main/java/com/alibaba/druid/sql/dialect/mysql/parser/MySqlExprParser.java
```diff
@@ -560,6 +560,37 @@ protected void parseIndexOptions(SQLIndexDefinition indexDefinition) {
...
+                    rightExp = primaryRest(rightExp);
+                } else {
+                    rightExp = bitOr();
+                }
```
wenshao, line 583: **[Suggestion]** The non-parenthesized `MEMBER OF` branch (`else { rightExp = bitOr(); }`) has no test coverage. — Concrete cost: if `bitOr()` produces an incorrect right-hand expression for certain inputs (e.g. compound expressions where precedence matters), no test catches the wrong AST or output. The parenthesized path calls `expr()` + `primaryRest()` while this path calls `bitOr()` — the asymmetry is unverified.

core/src/main/java/com/alibaba/druid/sql/dialect/mysql/parser/MySqlExprParser.java
```diff
@@ -560,6 +560,37 @@ protected void parseIndexOptions(SQLIndexDefinition indexDefinition) {
...
+
+                expr = new SQLBinaryOpExpr(expr, SQLBinaryOperator.MemberOf, rightExp, dbType);
+                return super.relationalRest(expr);
+            }
+            lexer.reset(mark);
+        }
```
wenshao, line 589: **[Suggestion]** The backtracking path — when `MEMBER` is used as a regular identifier (column/table name) NOT followed by `OF` — has no test coverage. `MEMBER` is not a reserved token, so any query like `SELECT member FROM t` enters this new `relationalRest()` override and must `lexer.reset(mark)` to backtrack. — Concrete cost: if a future change breaks the save/reset logic, these common queries fail to parse with no test to catch it.

wenshao (changes requested):

PR #6639, opened 2026-04-21
add Kingbase compatibility support

Angryberry:


Angryberry:
Hi maintainers,

The CLA check has passed, and there are no conflicts with the base branch now.

I would appreciate it if a maintainer could take a look at the pending workflow approval when convenient.

Thanks a lot!

core/src/main/java/com/alibaba/druid/pool/xa/DruidXADataSource.java
```diff
@@ -79,6 +79,8 @@ private XAConnection createPhysicalXAConnection(Connection physicalConn) throws
                 return MySqlUtils.createXAConnection(driver, physicalConn);
             case postgresql:
                 return PGUtils.createXAConnection(physicalConn);
+            case kingbase:
```
wenshao, line 82: **[Critical]** `KingbaseUtils.createXAConnection(url, user, password)` ignores the `physicalConn` parameter that the pool already created. Every other database type (oracle, mysql, postgresql, h2, jtds, dm, oceanbase) wraps or reuses `physicalConn`. This leaks the pool-managed connection and opens a second, independent physical connection outside the pool's control.

```suggestion
            case kingbase:
                return PGUtils.createXAConnection(physicalConn);
```

If the Kingbase XA driver requires a separate XADataSource (not wrapping an existing connection), the unused `physicalConn` must at minimum be properly handled by the caller to avoid connection leaks.

core/src/main/java/com/alibaba/druid/wall/spi/KingbaseWallProvider.java
```diff
@@ -0,0 +1,273 @@
...
+                    + "mode, using PostgreSQL as default");
+            return DbType.postgresql;
+        }
+
+        Connection conn = null;
+        Statement stmt = null;
+        ResultSet rs = null;
+
+        try {
+            // 获取原始连接（绕过 Filter 链，避免循环）
```
wenshao, line 100: **[Critical]** `queryCompatibilityMode()` opens a raw JDBC connection via `dataSource.getRawDriver().connect()` with **no connect timeout and no statement timeout**. If the Kingbase database is slow or unreachable, this blocks indefinitely on the first SQL check, potentially hanging the calling thread forever.

Suggested fix:
```java
stmt.setQueryTimeout(5); // 5-second timeout
```
Also consider setting a connect timeout via connection properties or `DriverManager.setLoginTimeout()`.

core/src/main/java/com/alibaba/druid/wall/spi/KingbaseWallProvider.java
```diff
@@ -0,0 +1,273 @@
...
+
+    /**
+     * 创建 SQL 解析器.
+     * @param sql SQL 语句
+     * @return SQL 解析器
+     */
+    @Override
+    public SQLStatementParser createParser(final String sql) {
+        DbType mode = getCompatibilityMode();
+        SQLParserFeature[] features =
```
wenshao, line 213: **[Critical]** When compat mode is MySQL, `MySqlStatementParser` is created with only `SQLParserFeature.EnableSQLBinaryOpExprGroup`. The real `MySqlWallProvider.createParser()` additionally includes `SQLParserFeature.StrictForWall`:

```java
// MySqlWallProvider.java (existing, correct)
return new MySqlStatementParser(sql,
    SQLParserFeature.EnableSQLBinaryOpExprGroup,
    SQLParserFeature.StrictForWall);
```

Without `StrictForWall`, MySQL conditional comments (e.g., `/*!50000 DROP TABLE users */`) are treated as harmless comments instead of executable SQL, allowing SQL injection through MySQL version-conditional comment syntax.

```suggestion
            case mysql:
                return new MySqlStatementParser(sql,
                        SQLParserFeature.EnableSQLBinaryOpExprGroup,
                        SQLParserFeature.StrictForWall);
```

core/src/main/java/com/alibaba/druid/wall/spi/KingbaseWallProvider.java
```diff
@@ -0,0 +1,273 @@
...
+    /** 默认配置目录. */
+    public static final String DEFAULT_CONFIG_DIR =
+            PGWallProvider.DEFAULT_CONFIG_DIR;
+
+    /** 缓存的数据库兼容模式. */
+    private volatile DbType compatMode;
+    /** 兼容模式锁. */
+    private final Object compatModeLock = new Object();
+    /** 数据源代理. */
+    private DataSourceProxy dataSource;
```
wenshao, line 52: **[Suggestion]** `KingbaseWallProvider` is the **only** `WallProvider` subclass that stores a `DataSourceProxy` field. All other providers (Oracle, MySQL, PG, SQLServer, DB2, SQLite, CK) are stateless SQL-analysis components with no coupling to the data source. This breaks the established contract.

Consider decoupling compat-mode detection from the provider — e.g., detect compat mode in `WallFilter.init()` and pass it as a constructor parameter, or make it configurable via `WallConfig`.

core/src/main/java/com/alibaba/druid/util/KingbaseUtils.java
```diff
@@ -0,0 +1,63 @@
...
+        SQLException sqlException = new SQLException("kingbase xa not supported: XADataSource class not found or incompatible");
+        if (lastError != null) {
+            sqlException.initCause(lastError);
+        }
+        throw sqlException;
+    }
+
+    private static void safeInvokeSetter(Object target, String methodName, Class<?> argType, Object argValue) {
+        try {
+            Method m = target.getClass().getMethod(methodName, argType);
```
wenshao, line 58: **[Suggestion]** `catch (Exception ignore)` silently swallows all exceptions including `SecurityException`. If a `SecurityManager` is active and denies the reflective setter call, the `SecurityException` is discarded and the code proceeds as if the setter succeeded — the XADataSource may then have uninitialized URL/user/password fields.

Suggested fix: catch `NoSuchMethodException` silently (expected for unsupported setter variants), let `SecurityException` propagate, and log other exceptions at DEBUG level.

core/src/main/java/com/alibaba/druid/util/KingbaseUtils.java
```diff
@@ -0,0 +1,63 @@
...
+            "com.kingbase.xa.KbXADataSource",
+            "kingbase8.xa.KbXADataSource",
+            "kingbase.xa.KbXADataSource"
+    };
+
+    public static XAConnection createXAConnection(String url, String user, String password) throws SQLException {
+        Exception lastError = null;
+        for (String className : CANDIDATE_XA_DS_CLASSNAMES) {
+            try {
+                Class<?> clazz = Class.forName(className);
```
wenshao, line 24: **[Suggestion]** `Class.forName(className)` is called on every `createXAConnection()` invocation, iterating up to 5 candidates each time. In XA transaction workloads, this adds unnecessary classloading overhead on every connection acquisition.

Cache the resolved `Class<?>` and relevant `Method` objects in a static volatile field after the first successful discovery.

core/src/main/java/com/alibaba/druid/util/KingbaseUtils.java
```diff
@@ -0,0 +1,63 @@
+/*
```
wenshao, line 1: **[Suggestion]** This file uses a single-line comment header (`/* Utility for Kingbase XA support via reflection */`) instead of the standard Apache License 2.0 header used in every other source file in the project (including `KingbaseWallProvider.java` in the same PR).

core/src/main/java/com/alibaba/druid/DbType.java
```diff
@@ -158,7 +158,11 @@ public static DbType of(String name) {
     }
 
     public static boolean isPostgreSQLDbStyle(DbType dbType) {
-        return dbType == DbType.postgresql || dbType == DbType.edb || dbType == DbType.greenplum || dbType == DbType.hologres;
+        return dbType == DbType.postgresql
+                || dbType == DbType.edb
+                || dbType == DbType.greenplum
```
wenshao, line 163: **[Suggestion]** `isPostgreSQLDbStyle()` now returns `true` for `kingbase`, but `KingbaseWallProvider.createParser()` can return `OracleStatementParser`, `MySqlStatementParser`, or `SQLServerStatementParser` depending on the runtime compat mode. Any caller that branches on `isPostgreSQLDbStyle(DbType.kingbase)` will assume PG semantics while the actual parser may be Oracle/MySQL/SQLServer.

Also, `JdbcUtils.isPgsqlDbType()` was NOT updated to include kingbase, creating an inconsistency between these two overlapping methods.

core/src/main/java/com/alibaba/druid/pool/DataSourceDisableException.java
```diff
@@ -20,15 +20,26 @@
 public class DataSourceDisableException extends SQLException {
     private static final long serialVersionUID = 1L;
 
+    /**
+     * 构造函数.
+     */
```
wenshao, line 25: **[Suggestion]** These are purely cosmetic changes (Javadoc + `final` parameters) unrelated to Kingbase compatibility support. Please revert or move to a separate cleanup PR to keep this PR focused.

core/src/main/java/com/alibaba/druid/pool/WrapperAdapter.java
```diff
@@ -18,18 +18,32 @@
 import java.sql.Wrapper;
 
 public class WrapperAdapter implements Wrapper {
+    /**
+     * 构造函数.
```
wenshao, line 22: **[Suggestion]** Same as above — cosmetic changes (Javadoc + `final` parameters) unrelated to Kingbase support. Please revert or move to a separate PR.

pom.xml
```diff
@@ -162,7 +162,7 @@
 							<goal>check</goal>
 						</goals>
 						<configuration>
-							<skip>false</skip>
+							<skip>${checkstyle.skip}</skip>
```
wenshao, line 165: **[Suggestion]** This checkstyle `skip` configuration change (`false` → `${checkstyle.skip}`) is unrelated to Kingbase support. It's a build-infrastructure change that should be reviewed and merged on its own merits in a separate PR.

wenshao (changes requested):
## Review Summary

**9 Critical / 10 Suggestion issues found.** This PR adds Kingbase compatibility support but has significant integration gaps and correctness concerns.

### Critical issues not mappable to a specific diff line:

**1. Binary `.class` file committed** — `KingBaseWallTest$MockDataSourceProxy.class` at the repository root is a compiled test artifact. Please remove it (`git rm`) and add `*.class` to `.gitignore`.

**2. Missing `kingbase` in multiple PG-family factory methods** — `SQLParserUtils` (4 methods: `createSQLStatementParser`, `createExprParser`, `createLexer`, `createSelectQueryBlock`), `SQLUtils` (`createOutputVisitor`, `createSchemaStatVisitor`), `JdbcUtils.isPgsqlDbType()`, and at least 7 additional switch blocks (`PagerUtils`, `SchemaRepository`, `SQLEvalVisitorUtils`, `ExportParameterVisitorUtils`, `SQLDeleteBuilderImpl`, `SQLUpdateBuilderImpl`, etc.) all group `postgresql/edb/greenplum` together but have no `case kingbase`. Most critically, `PagerUtils` will throw `UnsupportedOperationException` for kingbase queries, and `JdbcUtils.isPgsqlDbType()` contradicts `DbType.isPostgreSQLDbStyle()`.

**3. Zero test coverage** — No unit tests exist for `KingbaseWallProvider` (273 lines), `KingbaseUtils` (63 lines), or any of the new switch-case branches. The committed `.class` file suggests a test was written but only the compiled output was staged.

**4. SQL Wall bypass risk** — `KingbaseWallProvider` queries the database to determine which SQL parser dialect to use. A compromised or misconfigured database can report a different compat mode than it actually executes in, causing the WallFilter to parse SQL with the wrong dialect and potentially allowing SQL injection through the parser/execution semantic gap.

core/src/main/java/com/alibaba/druid/wall/spi/KingbaseWallProvider.java
```diff
@@ -0,0 +1,273 @@
...
+
+    /**
+     * 构造函数.
+     * @param config 配置
+     * @param dataSourceProxy 数据源代理
+     */
+    public KingbaseWallProvider(final WallConfig config,
+                                final DataSourceProxy dataSourceProxy) {
+        super(config, DbType.kingbase);
+        this.dataSource = dataSourceProxy;
```
wenshao, line 77: **[Critical]** `WallProvider` base class uses `DbType.kingbase` for SQL parameterization (`ParameterizedOutputVisitorUtils.parameterize(sql, dbType)`) and serialization (`SQLUtils.toSQLString`), but `createParser()`/`createWallVisitor()` return Oracle/MySQL/SQLServer implementations when compat mode is non-PG. This dialect mismatch causes parameterization to re-parse Oracle-syntax SQL with a generic parser and PG output visitor, which fails on Oracle syntax (ROWNUM, CONNECT BY, `(+)` joins). — Failure scenario: Kingbase in Oracle compat mode → parameterization fails → cache keys fall back to raw SQL with literals → same query template with different parameter values produces different cache entries → white/black list cache pollution; a blacklisted SQL pattern with different literals may evade the blacklist.

```suggestion
        super(config, DbType.kingbase);
        this.dataSource = dataSourceProxy;
        // TODO: override getMergedSqlNullableIfParameterizeError() or expose
        // effective DbType (compatMode) for parameterization/serialization
```

core/src/main/java/com/alibaba/druid/wall/spi/KingbaseWallProvider.java
```diff
@@ -0,0 +1,273 @@
...
+
+        Connection conn = null;
+        Statement stmt = null;
+        ResultSet rs = null;
+
+        try {
+            // 获取原始连接（绕过 Filter 链，避免循环）
+            conn = dataSource.getRawDriver().connect(
+                    dataSource.getRawJdbcUrl(),
+                    dataSource.getConnectProperties());
```
wenshao, line 103: **[Critical]** `queryCompatibilityMode()` opens a raw JDBC connection using `getConnectProperties()` which does NOT include username/password. Standard Druid configs set credentials via `setUsername()`/`setPassword()`, not in connect properties. Compare with `DruidAbstractDataSource.createPhysicalConnection()` which explicitly adds `user`/`password` to the properties. — Failure scenario: standard Druid config → `Driver.connect()` fails with auth error → catch block logs warning → returns `DbType.postgresql` → on a Kingbase instance in Oracle compat mode, all SQL is permanently parsed with PG parser (cached in `volatile compatMode`) → spurious wall violations or missed violations; only a restart can recover.

```suggestion
                Properties props = new Properties();
                if (dataSource.getConnectProperties() != null) {
                    props.putAll(dataSource.getConnectProperties());
                }
                if (dataSource.getUsername() != null) {
                    props.put("user", dataSource.getUsername());
                }
                if (dataSource.getPassword() != null) {
                    props.put("password", dataSource.getPassword());
                }
                conn = dataSource.getRawDriver().connect(
                        dataSource.getRawJdbcUrl(),
                        props);
```

core/src/main/java/com/alibaba/druid/util/KingbaseUtils.java
```diff
@@ -0,0 +1,63 @@
...
+                safeInvokeSetter(xaDs, "setUser", String.class, user);
+                safeInvokeSetter(xaDs, "setUsername", String.class, user);
+                safeInvokeSetter(xaDs, "setPassword", String.class, password);
+
+                // obtain XAConnection: prefer no-arg if properties are set on XADataSource
+                try {
+                    Method m = clazz.getMethod("getXAConnection");
+                    Object xaConn = m.invoke(xaDs);
+                    return (XAConnection) xaConn;
+                } catch (NoSuchMethodException ignore) {
```
wenshao, line 39: **[Critical]** The `catch (NoSuchMethodException)` fallback to `getXAConnection(user, password)` is structurally unreachable. `javax.sql.XADataSource` defines both `getXAConnection()` and `getXAConnection(String, String)`, so any compliant Kingbase driver class will have both methods — `clazz.getMethod("getXAConnection")` always succeeds. When the no-arg invocation fails (e.g., `InvocationTargetException` wrapping a "URL not set" SQLException), it is NOT a `NoSuchMethodException`, so it escapes to the outer `catch (Exception ex)`, skipping the two-arg fallback that passes credentials directly and would succeed. — Failure scenario: `safeInvokeSetter` silently fails to configure URL → no-arg `getXAConnection()` throws → outer catch advances to next candidate → all 5 candidates fail identically → `SQLException("kingbase xa not supported")` even though the two-arg path on the first class would have worked.

```suggestion
                XAConnection xaConn = null;
                try {
                    Method m = clazz.getMethod("getXAConnection");
                    xaConn = (XAConnection) m.invoke(xaDs);
                } catch (Exception noArgFailed) {
                    try {
                        Method m = clazz.getMethod("getXAConnection", String.class, String.class);
                        xaConn = (XAConnection) m.invoke(xaDs, user, password);
                    } catch (Exception twoArgFailed) {
                        throw twoArgFailed;
                    }
                }
                return xaConn;
```

core/src/main/java/com/alibaba/druid/util/KingbaseUtils.java
```diff
@@ -0,0 +1,63 @@
...
+                    // fallback: getXAConnection(user, password)
+                    Method m = clazz.getMethod("getXAConnection", String.class, String.class);
+                    Object xaConn = m.invoke(xaDs, user, password);
+                    return (XAConnection) xaConn;
+                }
+            } catch (Exception ex) {
+                lastError = ex;
+            }
```
wenshao, line 47: **[Suggestion]** When the first candidate class is found but `getXAConnection()` fails (e.g., auth failure), the `InvocationTargetException` is overwritten by subsequent `ClassNotFoundException`s from remaining candidates, masking the real error. — Concrete cost: final `SQLException` says "class not found or incompatible" with a `ClassNotFoundException` cause, sending the developer to investigate their classpath when the real problem is invalid credentials.

```suggestion
            } catch (Exception ex) {
                if (lastError == null || !(ex instanceof ClassNotFoundException)) {
                    lastError = ex;
                }
            }
```

core/src/main/java/com/alibaba/druid/wall/spi/KingbaseWallProvider.java
```diff
@@ -0,0 +1,273 @@
...
+    /**
+     * 创建 SQL 解析器.
+     * @param sql SQL 语句
+     * @return SQL 解析器
+     */
+    @Override
+    public SQLStatementParser createParser(final String sql) {
+        DbType mode = getCompatibilityMode();
+        SQLParserFeature[] features =
+                {SQLParserFeature.EnableSQLBinaryOpExprGroup};
```
wenshao, line 214: **[Suggestion]** `KingbaseWallProvider`'s three mode-dispatching methods (`createParser`, `createWallVisitor`, `createExportParameterVisitor`) — 12 branch paths across 4 compatibility modes — have no test coverage. The existing codebase pattern (e.g., `PGWallTest`) demonstrates wall providers can be tested without a live database. — Concrete cost: a regression in mode-to-parser mapping (e.g., `case oracle` accidentally returning `MySqlStatementParser`) would cause wrong dialect parsing, producing false-positive/negative wall violations with no test to catch it.

core/src/main/java/com/alibaba/druid/wall/spi/KingbaseWallProvider.java
```diff
@@ -0,0 +1,273 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+
+public class KingbaseWallProvider extends WallProvider {
+    /** 日志. */
+    private static final Log LOG =
+            LogFactory.getLog(KingbaseWallProvider.class);
+    /** 默认配置目录. */
+    public static final String DEFAULT_CONFIG_DIR =
+            PGWallProvider.DEFAULT_CONFIG_DIR;
```
wenshao, line 45: **[Suggestion]** Wall config deny lists are hardcoded to PostgreSQL (`DEFAULT_CONFIG_DIR = PGWallProvider.DEFAULT_CONFIG_DIR`) regardless of detected compatibility mode. — Concrete cost: on a Kingbase instance in Oracle compat mode, `WallConfig` loads `postgres/deny-table.txt` (~1 entry) instead of `oracle/deny-table.txt` (1352 entries including `DBA_USERS`, `V$SESSION`, `SYS.USER$`). `SELECT * FROM DBA_USERS` passes the wall check because the PG deny list has no match. Same gap applies to `deny-function.txt`, `deny-schema.txt`, and `deny-object.txt`.

core/src/main/java/com/alibaba/druid/sql/visitor/ParameterizedOutputVisitorUtils.java
```diff
@@ -429,6 +429,7 @@ public static ParameterizedVisitor createParameterizedOutputVisitor(StringBuilde
             case greenplum:
             case edb:
             case gaussdb:
+            case kingbase:
                 return new PGOutputVisitor(out, true);
```
wenshao, line 433: **[Suggestion]** The diff adds `case kingbase:` to the parameterized output visitor (PGOutputVisitor) but does not add a corresponding `case kingbase:` to `SQLParserUtils.createSQLStatementParser`. The wall's parameterization path parses Kingbase SQL with a generic `SQLStatementParser` whose `primaryColonColon()` throws `ParserException` on `::` type-casts (extremely common in PG/Kingbase SQL, e.g., `SELECT id::varchar FROM t`). — Concrete cost: parameterization fails for every PG-specific SQL → white list caching disabled → every such query becomes a hard wall check on every execution, a permanent performance degradation for the most common Kingbase SQL patterns.

wenshao (changes requested):
**[Critical]** Binary `.class` build artifact (`KingBaseWallTest$MockDataSourceProxy.class`) committed to repository root. The `.gitignore` has no `*.class` rule. Remove with `git rm` and add `*.class` to `.gitignore`. The presence of this compiled test class also strongly suggests a test source file (`KingBaseWallTest.java`) was written but accidentally excluded from the commit.

PR #6642, opened 2026-04-29
fix: SQLDeleteStatement.accept0 missing from/using children (#6563)

### LINKED ISSUE #6563: com.alibaba.druid.sql.ast.statement.SQLDeleteStatement缺少对from节点的accept
Image
参观者缺少对from节点的访问
【背景】对delete语句进行表名修改，from子查询语句里面表名修改

Image

【版本】1.2.27

daguimu:
## Problem

The base `SQLDeleteStatement.accept0` walked only `with` / `tableSource` / `where`, silently skipping the `from` and `using` sub-trees. As a result, dialects that fall back to the base implementation (DB2, Snowflake, etc.) cannot be fully traversed by visitors that rely on the default tree walk - any AST node reachable only via `DELETE ... FROM <ts>` or `DELETE ... USING <ts>` was invisible to AST consumers.

`getChildren()` had the same omission, and `setUsing()` failed to wire the parent pointer (`setFrom` already does), breaking parent-chain traversal for that branch.

Reported in #6563 (DELETE statement visitor cannot reach the FROM sub-tree when modifying table names).

## Root Cause

Compare with `MySqlDeleteStatement.accept0`, which correctly walks `with`, `tableSource`, `where`, `from`, `using`, `orderBy`, `limit`. The base class - used directly by Snowflake, DB2 and any future dialect that does not override - was missing `from` and `using`. Snowflake and PG/Oscar parsers populate the `using` slot via `setUsing()`, but no parent was assigned, so even reverse-traversal would not work.

## Fix

In `SQLDeleteStatement`:

- `accept0`: walk `from` and `using` in addition to the previously-visited children.
- `getChildren`: include `from` and `using`.
- `setUsing`: wire the parent pointer, mirroring `setFrom`.

No changes to dialect-specific subclasses - they already override `accept0` and were unaffected.

## Tests Added

`core/src/test/java/com/alibaba/druid/sql/issues/Issue6563.java`:

- `test_snowflake_delete_using_visited` - parses `DELETE FROM t1 USING t_using_only WHERE ...`, walks the AST with a generic identifier collector, and asserts `t_using_only` (referenced ONLY through the USING clause) is reached.
- `test_db2_delete_from_visited` - parses `DELETE FROM t FROM t_from_only WHERE ...` for DB2, asserts `t_from_only` (reachable only through the secondary FROM) is visited.
- `test_getChildren_includes_from_and_using` - structural test on the AST API.
- `test_setUsing_sets_parent` - verifies the parent pointer is wired so reverse traversal works.

All four tests fail on master (4 failures, no errors) and pass with the fix.

Full `mvn test` for the `core` module: 4222 tests, 0 failures, 0 errors.

Fixes #6563

### DIFF
--- a/core/src/test/java/com/alibaba/druid/sql/issues/Issue6563.java
+++ b/core/src/test/java/com/alibaba/druid/sql/issues/Issue6563.java
@@ -1,0 +1,94 @@
+package com.alibaba.druid.sql.issues;
+
+import com.alibaba.druid.DbType;
+import com.alibaba.druid.sql.ast.SQLObject;
+import com.alibaba.druid.sql.ast.SQLStatement;
+import com.alibaba.druid.sql.ast.expr.SQLIdentifierExpr;
+import com.alibaba.druid.sql.ast.statement.SQLDeleteStatement;
+import com.alibaba.druid.sql.ast.statement.SQLExprTableSource;
+import com.alibaba.druid.sql.parser.SQLParserUtils;
+import com.alibaba.druid.sql.parser.SQLStatementParser;
+import com.alibaba.druid.sql.visitor.SQLASTVisitorAdapter;
+import org.junit.jupiter.api.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * SQLDeleteStatement.accept0 / getChildren skipped the {@code from} and
+ * {@code using} children, so generic visitors that rely on the default
+ * tree walk never reached identifiers inside
+ * {@code DELETE ... FROM ...} or {@code DELETE ... USING ...}.
+ *
+ * @see <a href="https://github.com/alibaba/druid/issues/6563">#6563</a>
+ */
+public class Issue6563 {
+    /** Collects every SQLIdentifierExpr name seen during a default tree walk. */
+    private static class NameCollector extends SQLASTVisitorAdapter {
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/alibaba/druid/sql/issues/Issue6563.java
```diff
@@ -0,0 +1,94 @@
+package com.alibaba.druid.sql.issues;
```
wenshao, line 1: **[Critical]** Test file is outside the `bvt` package — surefire won't discover or run it. The surefire include pattern in `core/pom.xml` is `**/bvt/**/*.java`. This file is at `com.alibaba.druid.sql.issues`, which doesn't match.

Move to `com.alibaba.druid.bvt.sql.issues` package:

```java
package com.alibaba.druid.bvt.sql.issues;
```

and move the file to `core/src/test/java/com/alibaba/druid/bvt/sql/issues/Issue6563.java`.

The code changes to `SQLDeleteStatement.java` look correct — `accept0`, `getChildren`, and `setUsing` parent wiring are all properly implemented. Once the test is relocated, this is ready to merge.

_— Qwen Code /review_

wenshao (changes requested):

wenshao (changes requested):
**[Critical]** Existing blocker (comment 3487888922, verified, still stands): Test file Issue6563.java is at package com.alibaba.druid.sql.issues, but surefire include pattern is **/bvt/**/*.java — the test will not be discovered or run by mvn test. Must move to com.alibaba.druid.bvt.sql.issues. Failure scenario: mvn test -pl core → surefire skips Issue6563.java → 4 regression tests never execute → accept0/getChildren/setUsing regression undetected by CI.

PR #6649, opened 2026-06-19
fix: 修正 maxWaitThreadCount 阈值判断和错误信息

smileknife-pro:
1. 将 notEmptyWaitThreadCount > maxWaitThreadCount 改为 >=， 使 maxWaitThreadCount 阈值正确生效
2. 错误信息中使用 notEmptyWaitThreadCount 替代 lock.getQueueLength()， 因为后者返回的是 ReentrantLock 等待线程数，而非实际排队等待连接的线程数

### DIFF
--- a/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
+++ b/core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
@@ -1611,10 +1611,10 @@ public class DruidDataSource extends DruidAbstractDataSource
 
             try {
                 if (maxWaitThreadCount > 0
-                        && notEmptyWaitThreadCount > maxWaitThreadCount) {
+                        && notEmptyWaitThreadCount >= maxWaitThreadCount) {
                     connectErrorCountUpdater.incrementAndGet(this);
                     throw new SQLException("maxWaitThreadCount " + maxWaitThreadCount + ", current wait Thread count "
-                            + lock.getQueueLength());
+                            + notEmptyWaitThreadCount);
                 }
 
                 if (onFatalError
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/alibaba/druid/pool/DruidDataSource.java
```diff
@@ -1611,10 +1611,10 @@ private DruidPooledConnection getConnectionInternal(long maxWait) throws SQLExce
 
             try {
                 if (maxWaitThreadCount > 0
-                        && notEmptyWaitThreadCount > maxWaitThreadCount) {
+                        && notEmptyWaitThreadCount >= maxWaitThreadCount) {
```
wenshao, line 1614: **[Critical]** The `>` to `>=` change is semantically correct (properly limits waiting threads to exactly `maxWaitThreadCount`), but it breaks the existing test `DruidDataSourceTest_notEmptyWait2`.

That test sets `maxWaitThreadCount=10`, spawns exactly 10 threads that all enter the wait state (`notEmptyWaitThreadCount` becomes 10), then starts an 11th `errorThread` and asserts `assertEquals(0, maxWaitErrorCount.get())`. With `>=`, the 11th thread sees `10 >= 10 = true` and is rejected — `maxWaitErrorCount` becomes 1, failing the assertion.

The test needs to be updated to match the corrected boundary behavior. Additionally, no existing test deterministically verifies the exact boundary condition (`notEmptyWaitThreadCount == maxWaitThreadCount`), so adding a dedicated boundary test would prevent future regressions.

```suggestion
                if (maxWaitThreadCount > 0
                        && notEmptyWaitThreadCount >= maxWaitThreadCount) {
```

Suggested test fix in `DruidDataSourceTest_notEmptyWait2`:
```java
final int THREAD_COUNT = 9;  // was 10; leave one slot so errorThread hits the boundary
```
Then assert `assertEquals(1, maxWaitErrorCount.get())` to validate the new `>=` semantics.

wenshao (changes requested):

wenshao (approved):
LGTM ✅

`>` → `>=` 的 off-by-one 修复逻辑正确，`notEmptyWaitThreadCount` 在 lock 保护下读写无并发问题。错误信息改用 `notEmptyWaitThreadCount` 比 `lock.getQueueLength()` 更准确。测试验证了边界条件行为，所有测试通过。

_— Qwen Code /review_

PR #6658, opened 2026-06-21
test: add regression tests for DruidDataSource

amirdeljouyi:
Hey 👋

I added two regression tests for DruidDataSource around cloning behavior.

Impact on coverage:

* Covers the cloning logic that copies data source settings and mutable collections, including [covered lines](https://github.com/alibaba/druid/blob/master/core/src/main/java/com/alibaba/druid/pool/DruidAbstractDataSource.java#L2145-L2206).
* This adds about 2.65% line coverage for DruidDataSource.

Why:

While looking through the DruidDataSource implementation, I noticed this behavior wasn't directly verified by the existing test suite. These tests cover cloning configured data sources, verifying that core connection and pool settings are preserved while mutable collections remain independent between the original and clone.

I hope you find these tests useful. Please let me know if you have any questions or concerns.

Thanks,
Amir

### DIFF
--- a/core/src/test/java/com/alibaba/druid/pool/DruidDataSourceTest.java
+++ b/core/src/test/java/com/alibaba/druid/pool/DruidDataSourceTest.java
@@ -2,7 +2,12 @@ package com.alibaba.druid.pool;
 
 import org.junit.jupiter.api.Test;
 
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Properties;
+
 import static org.junit.jupiter.api.Assertions.*;
+
 public class DruidDataSourceTest {
     /**
      * 验证将mysql jdbc url中可能出现的密码信息全都掩码的效果，目前会出现的密码key名有password,password1,password2,password3,trustCertificateKeyStorePassword,clientCertificateKeyStorePassword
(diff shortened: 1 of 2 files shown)


core/src/test/java/com/alibaba/druid/pool/DruidDataSourceTest.java
```diff
@@ -83,4 +89,65 @@ public void test_sanitizedUrl() {
...
+
+        Properties connectProperties = new Properties();
+        connectProperties.setProperty("socketTimeout", "5000");
+        connectProperties.setProperty("connectTimeout", "3000");
+        dataSource.setConnectProperties(connectProperties);
+
+        List<String> filterClassNames = dataSource.getFilterClassNames();
+        assertEquals(1, filterClassNames.size());
+
+        return dataSource;
```
wenshao, line 155: **[Suggestion]** The helper sets `connectProperties` and `driverClassName` but `test_clone_copiesCoreConfiguration` doesn't verify they survive cloning. `Properties` is mutable — if `clone()` shares the reference instead of deep-copying, modifying one data source's connect properties would silently affect the other. Consider adding:
```java
assertEquals(dataSource.getDriverClassName(), clone.getDriverClassName());
assertNotSame(dataSource.getConnectProperties(), clone.getConnectProperties());
assertEquals(dataSource.getConnectProperties(), clone.getConnectProperties());
```
The `assertNotSame` on `connectProperties` is the critical one — it verifies the clone gets an independent copy.

core/src/test/java/com/alibaba/druid/pool/DruidDataSourceTest.java
```diff
@@ -83,4 +89,65 @@ public void test_sanitizedUrl() {
...
+        dataSource.setDefaultAutoCommit(false);
+        dataSource.setConnectionInitSqls(Arrays.asList("set names utf8mb4", "set sql_mode='STRICT_TRANS_TABLES'"));
+        dataSource.setFilters("stat");
+
+        Properties connectProperties = new Properties();
+        connectProperties.setProperty("socketTimeout", "5000");
+        connectProperties.setProperty("connectTimeout", "3000");
+        dataSource.setConnectProperties(connectProperties);
+
+        List<String> filterClassNames = dataSource.getFilterClassNames();
```
wenshao: **[Suggestion]** This `assertEquals` is inside the helper method `createCloneSource()`, not in a test method. Assertions in setup helpers can make test failures harder to diagnose (the failure stack trace points to the helper, not the test). Consider moving this assertion to the test methods that call `createCloneSource()`, or replacing it with a comment documenting the precondition:
```java
// setFilters("stat") should register exactly one filter class
List<String> filterClassNames = dataSource.getFilterClassNames();
if (filterClassNames.size() != 1) {
    throw new IllegalStateException("expected 1 filter, got " + filterClassNames.size());
}
```

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI still running.

Thanks for adding regression tests for `DruidDataSource` cloning behavior. The tests are well-structured and cover the key scenarios. A couple of suggestions below to strengthen the coverage.

core/src/test/java/com/alibaba/druid/pool/DruidDataSourceTest.java
amirdeljouyi: Thanks for the suggestion. I added the `driverClassName` and `connectProperties` assertions, and the `assertNotSame` check exposed that `connectProperties` was being shared. I updated `cloneTo` so the clone gets its own `Properties` copy.

wenshao (approved):
No issues found. LGTM! ✅

PR #6663, opened 2026-07-02
Fix oceanbase_oracle DbType mask corrupted by int sign extension

vasiliy-mikhailov:
### Problem

`DbType.oceanbase_oracle` is declared as:

```java
oceanbase_oracle(1 << 31),
```

`1 << 31` is evaluated in `int`, so it is `Integer.MIN_VALUE` (`0x80000000`). When that `int` is passed to the `DbType(long mask)` constructor it is widened with sign extension to `0xFFFFFFFF80000000L`, setting all of the upper 33 bits instead of just bit 31. Every other value at bit 32 and above already uses the `1L << n` form (`polardb(1L << 32)`, `ali_oracle(1L << 33)`, ...); `oceanbase_oracle` is the only one that shifts an `int`.

### Impact

`DbType.of(...)` OR-combines masks:

```java
value |= type.mask;
```

Because `oceanbase_oracle.mask` has all high bits set, any combination that includes it corrupts the result and produces false positives for mask-based matching against db types at bit 32 and above (polardb, ali_oracle, mock, ...).

### Fix

```diff
-    oceanbase_oracle(1 << 31),
+    oceanbase_oracle(1L << 31),
```

so the mask is the intended single bit `0x0000000080000000L`.

### Test

Adds `DbTypeMaskTest` (core module) covering the raw mask value, an `of(oceanbase_oracle, polardb)` combination, and a general "every non-zero mask is a single bit" invariant. All three assertions fail before the fix and pass after.

### DIFF
--- a/core/src/test/java/com/alibaba/druid/test/DbTypeMaskTest.java
+++ b/core/src/test/java/com/alibaba/druid/test/DbTypeMaskTest.java
@@ -1,0 +1,36 @@
+package com.alibaba.druid.test;
+
+import com.alibaba.druid.DbType;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class DbTypeMaskTest {
+    @Test
+    public void test_oceanbase_oracle_mask_no_sign_extension() {
+        long expectedMask = 1L << 31; // 0x0000000080000000L
+        assertEquals(expectedMask, DbType.oceanbase_oracle.mask,
+                "oceanbase_oracle mask should be a single clean bit at position 31, not sign-extended");
+    }
+
+    @Test
+    public void test_of_oceanbase_oracle_plus_polardb_no_corruption() {
+        long combined = DbType.of(DbType.oceanbase_oracle, DbType.polardb);
+        long expected = (1L << 31) | (1L << 32);
+        assertEquals(expected, combined,
+                "Combining oceanbase_oracle + polardb should not corrupt high bits due to sign extension");
+    }
+
+    @Test
+    public void test_all_masks_are_clean_single_bits() {
+        for (DbType type : DbType.values()) {
+            long mask = type.mask;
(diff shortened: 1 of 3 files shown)


wenshao (commented):

core/src/test/java/com/alibaba/druid/test/DbTypeMaskTest.java
```diff
@@ -0,0 +1,36 @@
+package com.alibaba.druid.test;
```
wenshao: **[Suggestion]** The new `DbTypeMaskTest` is in package `com.alibaba.druid.test`, but the core module's `maven-surefire-plugin` is configured with `<includes><include>**/bvt/**/*.java</include></includes>` (see `core/pom.xml`). An explicit `<includes>` overrides Surefire's default patterns, so a test outside the `**/bvt/**` tree is never selected for execution — this file compiles but is silently skipped by `mvn test` / `mvn clean install` and CI. (The `com.alibaba.druid.test` package is a manual/scratch area — e.g. `QuartzTest`, `AlibTest` — not wired into the automated suite.) — Failure scenario: the regression guard is inert; if this exact sign-extension bug is reintroduced later, CI stays green and it ships, and the PR's "tests fail before / pass after" claim cannot be validated by the build.

Suggested fix: move the test under the `bvt` tree so it matches the include pattern, e.g. `core/src/test/java/com/alibaba/druid/bvt/DbTypeMaskTest.java` with `package com.alibaba.druid.bvt;` (the project's established test convention).

vasiliy-mikhailov:
Good catch, thanks. Moved DbTypeMaskTest into the `com.alibaba.druid.bvt` package so it matches the surefire `**/bvt/**` include and actually runs in `mvn test` / CI (confirmed it executes, 3 tests). The test was silently skipped where it was.

wenshao (approved):
No issues found. LGTM! ✅

PR #6665, opened 2026-07-09
Decode with UTF-8 in ConfigTools.decrypt to match encrypt

vasiliy-mikhailov:
## What is wrong

`ConfigTools` **encrypts** passwords with **UTF-8** but **decrypts** with the **JVM default charset**. On any JVM whose default charset is not UTF-8 (Windows GBK, or a container started with `-Dfile.encoding=ISO-8859-1`), decrypting a password that contains non-ASCII characters returns garbage.

## Where (file `core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java`)

- **Line 179** (`encrypt`) encodes with UTF-8:
  ```java
  byte[] encryptedBytes = cipher.doFinal(plainText.getBytes("UTF-8"));
  ```
- **Line 146** (`decrypt`) decodes with the platform default charset (no charset argument):
  ```java
  return new String(plainBytes);      // uses Charset.defaultCharset()
  ```

The two sides disagree on charset, so the round trip is only correct when the default happens to be UTF-8.

## How it fails

Encrypt then decrypt a password with multibyte characters on a non-UTF-8 JVM:

```java
String pw = "密码P@sséü";
ConfigTools.decrypt(ConfigTools.encrypt(pw));   // does not equal pw
```

The bytes were written as UTF-8, but `decrypt` reads them back as (for example) ISO-8859-1.

## Test (`core/src/test/java/com/alibaba/druid/bvt/filter/config/ConfigToolsCharsetTest.java`)

```java
@Test
public void test_decrypt_roundtrips_multibyte_regardless_of_default_charset() throws Exception {
    String password = "密码P@sséü";
    String cipher = ConfigTools.encrypt(password);
    String plain  = ConfigTools.decrypt(cipher);
    assertEquals(password, plain,
        "decrypt must round-trip a UTF-8-encoded password even when the JVM default charset is not UTF-8");
}
```

Run with the default charset forced to a non-UTF-8 value (the condition under which the bug appears):

```
mvn test -Dtest=ConfigToolsCharsetTest -pl core -DargLine="-Dfile.encoding=ISO-8859-1"
```

Before the fix it fails:

```
org.opentest4j.AssertionFailedError: decrypt must round-trip a UTF-8-encoded password even when the JVM
default charset is not UTF-8 ==> expected: <密码P@sséü> but was: <å¯ç P@ssÃ©Ã¼>
```

## Fix

```diff
- return new String(plainBytes);
+ return new String(plainBytes, "UTF-8");
```

Decode with UTF-8 so `decrypt` matches `encrypt`.

## Verification

`mvn test -Dtest=ConfigToolsCharsetTest -pl core -DargLine="-Dfile.encoding=ISO-8859-1"` — **red before** (expected `密码P@sséü`, got `å¯ç P@ssÃ©Ã¼`), **green after** (`Tests run: 1, Failures: 0, BUILD SUCCESS`). On a default-UTF-8 JVM the round trip already works, which is why the test forces the non-UTF-8 condition.

## AI assistance disclosure

### DIFF
--- a/core/src/test/java/com/alibaba/druid/bvt/filter/config/ConfigToolsCharsetTest.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/filter/config/ConfigToolsCharsetTest.java
@@ -1,0 +1,19 @@
+package com.alibaba.druid.bvt.filter.config;
+
+import com.alibaba.druid.filter.config.ConfigTools;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ConfigToolsCharsetTest {
+    // A password with multibyte (non-ASCII) characters. encrypt() encodes with UTF-8,
+    // so decrypt() must also decode with UTF-8, independent of the JVM default charset.
+    @Test
+    public void test_decrypt_roundtrips_multibyte_regardless_of_default_charset() throws Exception {
+        String password = "密码P@sséü";   // 密码P@sséü
+        String cipher = ConfigTools.encrypt(password);
+        String plain = ConfigTools.decrypt(cipher);
+        assertEquals(password, plain,
+                "decrypt must round-trip a UTF-8-encoded password even when the JVM default charset is not UTF-8");
+    }
+}
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -68,6 +68,25 @@
 						<include>**/bvt/**/*.java</include>
 					</includes>
 				</configuration>
+				<executions>
+					<!-- Run the charset round-trip test under a non-UTF-8 default charset so it
+					     actually fails if decrypt() stops decoding as UTF-8 (it passes trivially
+					     when the JVM default is already UTF-8). -->
+					<execution>
+						<id>charset-tests</id>
+						<phase>test</phase>
+						<goals>
+							<goal>test</goal>
+						</goals>
+						<configuration>
+							<includes>
+								<include>**/bvt/filter/config/ConfigToolsCharsetTest.java</include>
+							</includes>
+							<argLine>-Dfile.encoding=ISO-8859-1</argLine>
+							<reportsDirectory>${project.build.directory}/surefire-reports-charset</reportsDirectory>
+						</configuration>
+					</execution>
+				</executions>
 			</plugin>
 			<plugin>
 				<artifactId>maven-source-plugin</artifactId>
--- a/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
+++ b/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
@@ -143,7 +143,7 @@ public class ConfigTools {
         byte[] cipherBytes = Base64.base64ToByteArray(cipherText);
         byte[] plainBytes = cipher.doFinal(cipherBytes);
 
-        return new String(plainBytes);
+        return new String(plainBytes, "UTF-8");
     }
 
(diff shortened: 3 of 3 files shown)


core/src/test/java/com/alibaba/druid/bvt/filter/config/ConfigToolsCharsetTest.java
```diff
@@ -0,0 +1,19 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ConfigToolsCharsetTest {
+    // A password with multibyte (non-ASCII) characters. encrypt() encodes with UTF-8,
+    // so decrypt() must also decode with UTF-8, independent of the JVM default charset.
+    @Test
+    public void test_decrypt_roundtrips_multibyte_regardless_of_default_charset() throws Exception {
+        String password = "密码P@sséü";   // 密码P@sséü
+        String cipher = ConfigTools.encrypt(password);
+        String plain = ConfigTools.decrypt(cipher);
```
wenshao, line 15: **[Suggestion]** The test never varies the JVM default charset, so it does not actually exercise the condition its name and assertion message claim (`regardless of default charset` / `even when the JVM default charset is not UTF-8`). — Failure scenario: the bug only manifests when the JVM default charset is non-UTF-8, but this test runs in the ambient charset. On every CI leg here (`ubuntu-latest` × Java 8/11/17/21, no `-Dfile.encoding`), the *unfixed* `new String(plainBytes)` also decodes as UTF-8, so the test passes with **or without** the fix. If a future change reverts `decrypt` back to `new String(plainBytes)`, CI stays green on all four legs and the regression this PR fixes ships undetected — the test guards nothing in the environment it actually runs in.

To genuinely gate the fix, exercise a non-UTF-8 charset. `Charset.defaultCharset()` is cached and cannot be reliably mutated at runtime, so fork a test JVM with a non-UTF-8 default — e.g. a dedicated surefire execution scoped to this test:

```xml
<execution>
  <id>configtools-non-utf8</id>
  <goals><goal>test</goal></goals>
  <configuration>
    <includes><include>**/ConfigToolsCharsetTest.java</include></includes>
    <argLine>-Dfile.encoding=ISO-8859-1</argLine>
  </configuration>
</execution>
```

Then the test fails before the fix and passes after. (Alternatively, soften the test name/message to describe a UTF-8 round-trip rather than default-charset independence.)

_— Qwen via Qwen Code /review_

wenshao (commented):

_— Qwen via Qwen Code /review_

vasiliy-mikhailov:
You're right, the round-trip passed on the UTF-8 CI legs whether or not the fix was present, so it wasn't actually gating anything.

I added a dedicated `charset-tests` surefire execution that forks the test JVM with `-Dfile.encoding=ISO-8859-1`, so `ConfigToolsCharsetTest` now runs under a non-UTF-8 default charset. Verified by execution on JDK 8: with `decrypt()` on `new String(plainBytes)` it fails (the multibyte password comes back garbled), and with `new String(plainBytes, "UTF-8")` it passes. So the test now genuinely fails before the fix and passes after.

wenshao (approved):
No issues found. LGTM! ✅

<details>
<summary>中文说明</summary>

未发现问题。LGTM！✅

</details>

wenshao (commented):
⚠️ Downgraded from Approve to Comment: CI still running. Reviewed.

<details>
<summary>中文说明</summary>

⚠️ 已从批准降级为评论：CI still running。 已审查。

</details>

PR #6666, opened 2026-07-09
Compare filter class names case-sensitively in FilterManager.existsFilter

vasiliy-mikhailov:
## What is wrong

`FilterManager.existsFilter(...)` checks whether a filter class is already registered by comparing fully-qualified class names with `equalsIgnoreCase`. Java class names are **case-sensitive** (`com.foo.MyFilter` and `com.foo.myfilter` are different classes), so two distinct filter classes whose names differ only in case are treated as the same one, and the second is silently skipped instead of being registered.

## Where (file `core/src/main/java/com/alibaba/druid/filter/FilterManager.java`)

- **Line 161** — the private dedup check:
  ```java
  private static boolean existsFilter(List<Filter> filterList, String filterClassName) {
  ```
- **Line 164** — compares class names case-insensitively:
  ```java
  if (itemFilterClassName.equalsIgnoreCase(filterClassName)) {   // class names are case-sensitive
  ```

`existsFilter` gates registration at lines 108 and 143, so a false positive here drops a real filter.

## How it fails

```java
List<Filter> filters = new ArrayList<>();
filters.add(new MyFilter());                       // com.foo.MyFilter
existsFilter(filters, "com.foo.myfilter");         // returns TRUE -> a different class is treated as present
```

## Test (`core/src/test/java/com/alibaba/druid/bvt/filter/FilterManagerTest.java`)

```java
@Test
public void test_existsFilter_caseSensitive() throws Exception {
    List<Filter> filterList = new ArrayList<>();
    filterList.add(new CaseTestFilter());
    String exactName = CaseTestFilter.class.getName();
    String lowerName = exactName.toLowerCase();
    Method existsFilter = FilterManager.class.getDeclaredMethod("existsFilter", List.class, String.class);
    existsFilter.setAccessible(true);
    assertTrue((Boolean) existsFilter.invoke(null, filterList, exactName));   // exact match
    assertFalse((Boolean) existsFilter.invoke(null, filterList, lowerName),   // different class, different name
            "existsFilter should be case-sensitive; class names that differ only in case are different classes");
}
```

Before the fix it fails:

```
org.opentest4j.AssertionFailedError: existsFilter should be case-sensitive; class names that differ
only in case are different classes ==> expected: <false> but was: <true>
```

## Fix

```diff
- if (itemFilterClassName.equalsIgnoreCase(filterClassName)) {
+ if (itemFilterClassName.equals(filterClassName)) {
```

## Verification

`mvn test -Dtest=FilterManagerTest#test_existsFilter_caseSensitive -pl core` — **red before** (`expected: <false> but was: <true>`), **green after** (`Tests run: 1, Failures: 0`).

## AI assistance disclosure

### DIFF
--- a/core/src/test/java/com/alibaba/druid/bvt/filter/FilterManagerTest.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/filter/FilterManagerTest.java
@@ -56,4 +56,22 @@ public class FilterManagerTest {
             throw new RuntimeException();
         }
     }
+
+    @org.junit.jupiter.api.Test
+    public void test_existsFilter_caseSensitive() throws Exception {
+        java.util.List<com.alibaba.druid.filter.Filter> filterList = new java.util.ArrayList<>();
+        filterList.add(new CaseTestFilter());
+        String exactName = CaseTestFilter.class.getName();
+        String lowerName = exactName.toLowerCase();
+        java.lang.reflect.Method existsFilter = com.alibaba.druid.filter.FilterManager.class
+                .getDeclaredMethod("existsFilter", java.util.List.class, String.class);
+        existsFilter.setAccessible(true);
+        org.junit.jupiter.api.Assertions.assertTrue((Boolean) existsFilter.invoke(null, filterList, exactName));
+        org.junit.jupiter.api.Assertions.assertFalse((Boolean) existsFilter.invoke(null, filterList, lowerName),
+                "existsFilter should be case-sensitive; class names that differ only in case are different classes");
+    }
+
+    public static class CaseTestFilter extends com.alibaba.druid.filter.FilterAdapter {
+    }
+
 }
--- a/core/src/test/java/com/alibaba/druid/bvt/filter/FilterManagerTest.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/filter/FilterManagerTest.java
@@ -5,8 +5,10 @@ import com.alibaba.druid.filter.FilterAdapter;
 import com.alibaba.druid.filter.FilterManager;
 import org.junit.jupiter.api.Test;
 
+import java.lang.reflect.Method;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.*;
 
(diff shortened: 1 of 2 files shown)


wenshao (approved):
No issues found. LGTM! ✅

wenshao (commented):

core/src/test/java/com/alibaba/druid/bvt/filter/FilterManagerTest.java
```diff
@@ -56,4 +56,22 @@ public ErrorFilter() {
             throw new RuntimeException();
         }
     }
+
+    @org.junit.jupiter.api.Test
+    public void test_existsFilter_caseSensitive() throws Exception {
```
wenshao: **[Suggestion]** The new test uses fully-qualified class names (`@org.junit.jupiter.api.Test`, `java.util.List<com.alibaba.druid.filter.Filter>`, `new java.util.ArrayList<>()`, `com.alibaba.druid.filter.FilterManager.class`, `org.junit.jupiter.api.Assertions.assertTrue/assertFalse`, `com.alibaba.druid.filter.FilterAdapter`) even though the file already imports all of these types. — Concrete cost: every existing test in this file uses the short forms (`@Test`, `assertNotNull`, `FilterAdapter`, `ArrayList`); the inconsistency makes the new test harder to read and immune to IDE rename-refactoring.

```suggestion
    @Test
    public void test_existsFilter_caseSensitive() throws Exception {
        List<Filter> filterList = new ArrayList<>();
        filterList.add(new CaseTestFilter());

        String exactName = CaseTestFilter.class.getName();
        String lowerName = exactName.toLowerCase();

        java.lang.reflect.Method existsFilter = FilterManager.class
                .getDeclaredMethod("existsFilter", List.class, String.class);
        existsFilter.setAccessible(true);

        assertTrue((Boolean) existsFilter.invoke(null, filterList, exactName));
        assertFalse((Boolean) existsFilter.invoke(null, filterList, lowerName),
                "existsFilter should be case-sensitive; class names that differ only in case are different classes");
    }
```

vasiliy-mikhailov: Good catch — that method slipped in fully-qualified. Switched it (and `CaseTestFilter`) to the short forms to match the rest of the file, and added the two imports that were genuinely missing (`java.util.List`, `java.lang.reflect.Method`). Purely cosmetic: `FilterManagerTest` still runs 4/4 green, and I re-checked that it goes red against the old `equalsIgnoreCase`, so the regression test still bites.

PR #6667, opened 2026-07-09
Print usage instead of ArrayIndexOutOfBoundsException in ConfigTools.main with no args

vasiliy-mikhailov:
## What is wrong

`ConfigTools.main` reads `args[0]` without checking `args.length`, so running the tool with no arguments throws `ArrayIndexOutOfBoundsException` instead of telling the user how to use it.

## Where (file `core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java`)

- **Line 42-43** — `main` uses `args[0]` immediately:
  ```java
  public static void main(String[] args) throws Exception {
      String password = args[0];      // ArrayIndexOutOfBoundsException when args is empty
  ```

## How it fails

```
$ java -cp druid.jar com.alibaba.druid.filter.config.ConfigTools
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 0
    at com.alibaba.druid.filter.config.ConfigTools.main(ConfigTools.java:43)
```

## Test (`core/src/test/java/com/alibaba/druid/bvt/filter/config/ConfigToolsMainTest.java`)

```java
@Test
public void main_withNoArgs_printsUsage_notAIOOBE() {
    Assertions.assertDoesNotThrow(() -> ConfigTools.main(new String[0]),
            "main() with no args should print usage, not throw ArrayIndexOutOfBoundsException");
}
```

Before the fix it fails:

```
org.opentest4j.AssertionFailedError: ... ==> Unexpected exception thrown:
java.lang.ArrayIndexOutOfBoundsException: 0
```

## Fix

```diff
 public static void main(String[] args) throws Exception {
+    if (args.length == 0) {
+        System.out.println("Usage: ConfigTools <password>");
+        return;
+    }
     String password = args[0];
```

## Verification

`mvn test -Dtest=ConfigToolsMainTest -pl core` — **red before** (`ArrayIndexOutOfBoundsException: 0`), **green after** (prints `Usage: ConfigTools <password>`, `Tests run: 1, Failures: 0`).

## AI assistance disclosure

### DIFF
--- a/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
+++ b/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
@@ -40,6 +40,10 @@ public class ConfigTools {
     public static final String DEFAULT_PUBLIC_KEY_STRING = "MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAKHGwq7q2RmwuRgKxBypQHw0mYu4BQZ3eMsTrdK8E6igRcxsobUC7uT0SoxIjl1WveWniCASejoQtn/BY6hVKWsCAwEAAQ==";
 
     public static void main(String[] args) throws Exception {
+        if (args.length == 0) {
+            System.out.println("Usage: ConfigTools <password>");
+            return;
+        }
         String password = args[0];
         String[] arr = genKeyPair(512);
         System.out.println("privateKey:" + arr[0]);
--- a/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
+++ b/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
@@ -41,7 +41,7 @@ public class ConfigTools {
 
     public static void main(String[] args) throws Exception {
         if (args.length == 0) {
-            System.out.println("Usage: ConfigTools <password>");
+            System.err.println("Usage: ConfigTools <password>");
             return;
         }
         String password = args[0];
--- a/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
+++ b/core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
@@ -40,15 +40,26 @@ public class ConfigTools {
     public static final String DEFAULT_PUBLIC_KEY_STRING = "MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAKHGwq7q2RmwuRgKxBypQHw0mYu4BQZ3eMsTrdK8E6igRcxsobUC7uT0SoxIjl1WveWniCASejoQtn/BY6hVKWsCAwEAAQ==";
 
     public static void main(String[] args) throws Exception {
+        int status = run(args);
+        if (status != 0) {
+            System.exit(status);
+        }
+    }
+
+    /**
+     * @return 0 when the keys were generated, 1 when the required password argument is missing
+     */
+    public static int run(String[] args) throws Exception {
         if (args.length == 0) {
             System.err.println("Usage: ConfigTools <password>");
-            return;
+            return 1;
         }
         String password = args[0];
         String[] arr = genKeyPair(512);
         System.out.println("privateKey:" + arr[0]);
         System.out.println("publicKey:" + arr[1]);
         System.out.println("password:" + encrypt(arr[0], password));
+        return 0;
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
```diff
@@ -40,6 +40,10 @@ public class ConfigTools {
     public static final String DEFAULT_PUBLIC_KEY_STRING = "MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAKHGwq7q2RmwuRgKxBypQHw0mYu4BQZ3eMsTrdK8E6igRcxsobUC7uT0SoxIjl1WveWniCASejoQtn/BY6hVKWsCAwEAAQ==";
 
     public static void main(String[] args) throws Exception {
+        if (args.length == 0) {
+            System.out.println("Usage: ConfigTools <password>");
```
wenshao: **[Suggestion]** This usage message goes to `System.out` — the same stream as the tool's real output (`privateKey`/`publicKey`/`password`). Diagnostics conventionally belong on `System.err`. Failure scenario: running `java ... ConfigTools > keys.txt` with an empty/unset password writes the `Usage:` line into `keys.txt` instead of surfacing it on the terminal, contaminating captured program output.

Related: the new `return` makes the no-args path exit 0, where the old `ArrayIndexOutOfBoundsException` exited non-zero — a script/CI step checking `$?` would now see success (with no key output) and miss the misconfiguration. If a non-zero exit is wanted, note that `System.exit(...)` here would terminate the surefire fork the new `assertDoesNotThrow` test runs in; the testable shape is a `static int run(String[])` that `main` wraps with `System.exit(run(args))`, which the test can assert on directly.

```suggestion
            System.err.println("Usage: ConfigTools <password>");
```

wenshao (commented):

vasiliy-mikhailov:
Good point, thanks. Moved the usage message to `System.err` so it no longer mixes with the tool's real output on `System.out`. The test now captures `System.err` and asserts the usage line is written there (fails before the change, passes after).

core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
```diff
@@ -40,6 +40,10 @@ public class ConfigTools {
     public static final String DEFAULT_PUBLIC_KEY_STRING = "MFwwDQYJKoZIhvcNAQEBBQADSwAwSAJBAKHGwq7q2RmwuRgKxBypQHw0mYu4BQZ3eMsTrdK8E6igRcxsobUC7uT0SoxIjl1WveWniCASejoQtn/BY6hVKWsCAwEAAQ==";
 
     public static void main(String[] args) throws Exception {
+        if (args.length == 0) {
+            System.err.println("Usage: ConfigTools <password>");
+            return;
```
wenshao: **[Suggestion]** Missing-argument invocation exits with code 0, indistinguishable from success. Before this PR, running `ConfigTools` with no arguments threw `ArrayIndexOutOfBoundsException` (non-zero exit). After this change, it prints usage to stderr and returns normally (exit 0), so a shell chain like `java -cp druid.jar com.alibaba.druid.filter.config.ConfigTools && deploy.sh` would silently proceed to `deploy.sh` as if key generation succeeded.

Concrete cost: Unix convention is that usage errors exit non-zero (typically 1). Scripts and CI steps using `&&` or `set -e` will not stop on this error.

```suggestion
        System.err.println("Usage: ConfigTools <password>");
        System.exit(1);
```

Note: if `System.exit(1)` is adopted, the test would need to be adjusted to a process-based or SecurityManager-based approach.

wenshao (commented):

core/src/main/java/com/alibaba/druid/filter/config/ConfigTools.java
vasiliy-mikhailov: Agreed — exit 0 on a usage error is wrong. I did not use `System.exit(1)` directly in `main`, though: surefire runs tests in a forked JVM, so a test calling `main()` would terminate the fork, and the SecurityManager alternative is deprecated for removal. Instead I extracted `run(String[])` returning a status code; `main` calls `System.exit` only when it is non-zero, so a successful run still returns normally. Verified end to end: with no args it exits 1 and `ConfigTools && deploy.sh` now aborts the chain, while `ConfigTools <password>` still exits 0.

PR #6674, opened 2026-07-21
fix(odps): preserve range-clustered bucket count

ygyydds:
## Summary

- preserve the bucket count for ODPS `RANGE CLUSTERED BY ... INTO n BUCKETS`
- add a BVT regression test for both the AST value and SQL round trip

## Root cause

`OdpsCreateTableParser` stored the range-clustered bucket count in `intoBuckets`, while the public `getBuckets()` API and `OdpsOutputVisitor` read `buckets`. This field mismatch made `getBuckets()` return `0` and silently removed the `INTO n BUCKETS` clause when serializing the parsed AST.

The parser now uses `setBuckets()`, matching the existing hash-clustered path and the output visitor.

## Reproduction

```sql
CREATE TABLE t (id BIGINT) RANGE CLUSTERED BY (id) INTO 1024 BUCKETS;
```

Before this change:

- `getBuckets()` returned `0`
- `SQLUtils.toOdpsString()` omitted `INTO 1024 BUCKETS`

After this change, the AST reports `1024` and the clause survives the SQL round trip.

## Testing

- `mvn -pl core "-Dtest=com.alibaba.druid.bvt.sql.odps.OdpsCreateTableTest4" test`
- `mvn -pl core "-Dtest=com/alibaba/druid/bvt/sql/odps/*" test` (225 tests)
- `mvn -pl core test` (4,357 tests, 0 failures, 0 errors, 1 skipped)

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/odps/parser/OdpsCreateTableParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/odps/parser/OdpsCreateTableParser.java
@@ -18,7 +18,6 @@ package com.alibaba.druid.sql.dialect.odps.parser;
 import com.alibaba.druid.sql.ast.ClusteringType;
 import com.alibaba.druid.sql.ast.SQLExpr;
 import com.alibaba.druid.sql.ast.SQLName;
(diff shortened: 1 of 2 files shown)


wenshao (approved):
No issues found. LGTM! ✅

PR #6675, opened 2026-07-23
Fix: GaussDB parser error on "ON DUPLICATE KEY UPDATE NOTHING"

Dra-kt:
Adds support for the `ON DUPLICATE KEY UPDATE NOTHING` clause in GaussDB SQL parsing.

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/gaussdb/parser/GaussDbStatementParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/gaussdb/parser/GaussDbStatementParser.java
@@ -158,8 +158,14 @@ public class GaussDbStatementParser extends PGSQLStatementParser {
                 accept(Token.UPDATE);
 
                 if (lexer.identifierEquals(FnvHash.Constants.NOTHING)) {
+                    int mark = lexer.mark();
                     lexer.nextToken();
-                    stmt.setDuplicateKeyUpdateNothing(true);
+                    if (lexer.token() == Token.EQ) {
+                        lexer.reset(mark);
+                    } else {
+                        stmt.setDuplicateKeyUpdateNothing(true);
+                    }
+                }
                 } else {
                     List<SQLExpr> duplicateKeyUpdate = stmt.getDuplicateKeyUpdate();
                     for (; ; ) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/alibaba/druid/sql/dialect/gaussdb/parser/GaussDbStatementParser.java
```diff
@@ -157,26 +157,31 @@ public PGInsertStatement parseInsert() {
...
-                    accept(Token.EQ);
-                    SQLExpr value;
-                    try {
-                        value = this.exprParser.expr();
-                    } catch (EOFParserException e) {
-                        throw new ParserException("EOF, " + name + "=", e);
-                    }
+                if (lexer.identifierEquals(FnvHash.Constants.NOTHING)) {
+                    lexer.nextToken();
+                    stmt.setDuplicateKeyUpdateNothing(true);
```
wenshao: **[Suggestion]** `NOTHING` is matched greedily without look-ahead, so a column named `nothing` in an assignment (`ON DUPLICATE KEY UPDATE nothing = 1`) is misparsed as the `NOTHING` keyword. `NOTHING` is not a reserved keyword in the GaussDB/PG lexer, so `identifierEquals(FnvHash.Constants.NOTHING)` matches it as a regular identifier. — Failure scenario: input `INSERT INTO t (a) VALUES (1) ON DUPLICATE KEY UPDATE nothing = 1` → the parser enters the `NOTHING` branch, consumes `nothing`, sets `duplicateKeyUpdateNothing = true`, and the remaining `= 1` tokens cause a `ParserException`. Note: the existing `ON CONFLICT DO NOTHING` path is unambiguous because `DO` precedes the keyword; here `UPDATE` is the same position where column assignments begin.

```suggestion
                if (lexer.identifierEquals(FnvHash.Constants.NOTHING)) {
                    int mark = lexer.mark();
                    lexer.nextToken();
                    if (lexer.token() == Token.EQ) {
                        lexer.reset(mark);
                    } else {
                        stmt.setDuplicateKeyUpdateNothing(true);
                    }
                }
```

wenshao (commented):

PR #6684, opened 2026-08-01
fix(ast): align SQLObject.setSource parameter names with impl and callers

### LINKED ISSUE #6311: [BUG] `SQLObject#setSource` parameters are switched
### Database Type

-

### Database Version

-

### Druid Version

c83288ed3b466b446951aed592fc5785313b13f9

### JDK Version

-

### Error SQL

-

### Testcase Code

The class `com.alibaba.druid.sql.ast.SQLObject` currently defines the method `setSource` like this:
```java
setSource(int column, int line)
```

However, the implementation in `SQLObjectImpl` and also all callers seem to provide `(line, column)` (switched parameters).

It would therefore be good to either switch the parameters in `SQLObject#setSource` to be `(int line, int column)` as well (and to match the implementation and usage), or to adjust `SQLObjectImpl` and all callers.

Please double-check in case I overlooked something.

I will probably not submit a pull request for this.

### Stacktrace Info

_No response_

### Error Info

_No response_

fudianchn:
## What

The `SQLObject.setSource` interface declaration now names its parameters `(int line, int column)`, matching the implementation in `SQLObjectImpl` and the argument order used by every caller.

## Why

The interface declared the parameters in the swapped order:

```java
// SQLObject.java (before)
default void setSource(int column, int line) { }
```

but the implementation and all 20+ callers use `(line, column)`:

```java
// SQLObjectImpl.java
public void setSource(int line, int column) {
    this.sourceLine = line;
    this.sourceColumn = column;
}

// callers, e.g. Lexer.java, SQLExprParser.java, OdpsExprParser.java
expr.setSource(line, column);
expr.setSource(sourceLine, sourceColumn);
```

The runtime behavior was correct — parameter names are not part of the bytecode, so positional calls were unaffected — but the misleading signature contradicted the implementation and every caller, inviting future positional mistakes (especially for anyone relying on IDE parameter hints or Javadoc).

## Root cause

`SQLObject.java:32` had the parameter names in the opposite order from `SQLObjectImpl.java:302` and all callers.

## How

Rename the interface parameters to `(int line, int column)`. No type/order change, no behavioral change, no caller change.

## Testing

- New test `SQLObjectSetSourceTest.setSource_lineFirstColumnSecond` pins the semantics: `setSource(2, 5)` yields `getSourceLine() == 2` and `getSourceColumn() == 5`, so a future swap is caught.
- Regression: `SourceLocationTest` → `Tests run: 1, Failures: 0`.

Fixes #6311

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/ast/SQLObject.java
+++ b/core/src/main/java/com/alibaba/druid/sql/ast/SQLObject.java
@@ -29,7 +29,7 @@ public interface SQLObject {
         return 0;
     }
 
-    default void setSource(int column, int line) {
+    default void setSource(int line, int column) {
     }
 
     void accept(SQLASTVisitor visitor);
--- a/core/src/test/java/com/alibaba/druid/bvt/sql/SQLObjectSetSourceTest.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/sql/SQLObjectSetSourceTest.java
@@ -1,0 +1,25 @@
+package com.alibaba.druid.bvt.sql;
+
+import com.alibaba.druid.sql.ast.expr.SQLCharExpr;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Locks in the parameter order of {@code SQLObject.setSource(int line, int column)}.
+ * See issue #6311: the interface declaration previously named the parameters in the swapped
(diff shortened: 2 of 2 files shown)


fudianchn:
Closing to reduce the Druid open-PR backlog I created by opening too many narrow parsing PRs at once. SQLObject.setSource parameter-name alignment (refactor) is a narrow grammar addition; I'll revisit it later as part of a grouped pass rather than keep this standalone.

PR #6685, opened 2026-08-01
fix(oracle): parse PL/SQL cursor RETURN clause in forward declarations

### LINKED ISSUE #5680: Cannot parse Oracle PL/SQL explicit cursor RETURN statement
```sql
DECLARE
  CURSOR c1 RETURN departments%ROWTYPE; 
...
```

Exception:
```
com.alibaba.druid.sql.parser.ParserException: syntax error, error in :'CURSOR c1 RETURN departments%ROWTYPE; CU', expect IS, actual RETURN, pos 25, line 1, column 20, token RETURN

```

fudianchn:
## What

The Oracle PL/SQL parser now accepts explicit cursor forward declarations with a `RETURN` clause, e.g. `CURSOR c1 RETURN departments%ROWTYPE;`, instead of failing with `expect IS, actual RETURN`.

## Why

A PL/SQL cursor spec may declare a return row type without a body (forward declaration):

```sql
DECLARE
  CURSOR c1 RETURN departments%ROWTYPE;
BEGIN
  NULL;
END;
```

The parser's cursor branch unconditionally `accept(Token.IS)` after the cursor name (and optional parameter list), expecting `IS SELECT ...`. When the spec carried a `RETURN` clause instead, it raised:

```
ParserException: syntax error, ... expect IS, actual RETURN, pos 25, line 1, column 20, token RETURN
```

`RETURN` is also legal *before* the `IS SELECT` body, so the fix needs to handle both a RETURN-only forward declaration and a RETURN + IS SELECT cursor.

## Root cause

`OracleStatementParser.java:2038` called `accept(Token.IS)` unconditionally after parsing the cursor name/parameters, with no branch for the `RETURN` clause that PL/SQL cursor specs allow.

## How

In `parserParameters`'s cursor branch, after the optional `(params)`:
- If the next token is `RETURN`, consume it and parse a data type into a new `SQLParameter.returnDataType` field.
- Make `IS SELECT` optional: only consume `IS` and the select body when a `RETURN`-only forward declaration is not what we have (i.e. when the next token is `IS`).

`SQLParameter` gains a `returnDataType` field (getter/setter, clone, and visitor child traversal), mirroring the existing `returnDataType` on `OracleCreateTypeStatement`. The output visitor renders `RETURN <type>` between the cursor name and the optional `IS SELECT`.

## Testing

- New test `OracleCursorReturnTest` (3 cases), all fail before the fix with `expect IS, actual RETURN`, all pass after:
  - `cursorWithReturnForwardDeclaration` — `CURSOR c1 RETURN departments%ROWTYPE;` parses and round-trips.
  - `cursorWithReturnAndSelect` — `CURSOR c1 RETURN departments%ROWTYPE IS SELECT ...` parses and the return type is captured on the AST.
  - `cursorWithoutReturnStillWorks` — regression guard: the existing `CURSOR c1 IS SELECT ...` form still parses.
- `./mvnw -pl core test-compile` → `BUILD SUCCESS`.

Before the fix:
```
DECLARE CURSOR c1 RETURN departments%ROWTYPE; BEGIN NULL; END;
  → ParserException: expect IS, actual RETURN
```

After the fix:
```
DECLARE CURSOR c1 RETURN departments%ROWTYPE; BEGIN NULL; END;
  → parses; round-trips to the canonical form with RETURN preserved
```

Fixes #5680

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/ast/SQLParameter.java
+++ b/core/src/main/java/com/alibaba/druid/sql/ast/SQLParameter.java
@@ -30,6 +30,7 @@ public final class SQLParameter extends SQLObjectImpl implements SQLObjectWithDa
     private boolean isNotNull;
     private SQLName cursorName;
     private final List<SQLParameter> cursorParameters = new ArrayList<>();
+    private SQLDataType returnDataType;
     private boolean order;
     private boolean map;
     private boolean member;
@@ -99,6 +100,7 @@ public final class SQLParameter extends SQLObjectImpl implements SQLObjectWithDa
             acceptChild(visitor, name);
             acceptChild(visitor, dataType);
             acceptChild(visitor, defaultValue);
+            acceptChild(visitor, returnDataType);
         }
         visitor.endVisit(this);
     }
@@ -149,6 +151,17 @@ public final class SQLParameter extends SQLObjectImpl implements SQLObjectWithDa
         this.cursorName = cursorName;
     }
 
+    public SQLDataType getReturnDataType() {
+        return returnDataType;
+    }
+
+    public void setReturnDataType(SQLDataType returnDataType) {
+        if (returnDataType != null) {
+            returnDataType.setParent(this);
+        }
+        this.returnDataType = returnDataType;
+    }
+
     public SQLParameter clone() {
         SQLParameter x = new SQLParameter();
         if (name != null) {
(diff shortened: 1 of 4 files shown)


fudianchn:
Closing for now to cut the backlog I created by opening too many Druid PRs at once. PL/SQL cursor RETURN clause in forward declarations is a fuller change; I'll re-file it once the open count is down so it gets focused review.

PR #6687, opened 2026-08-01
fix(derby): render row-limiting clause as FETCH FIRST instead of LIMIT

### LINKED ISSUE #3612: SQLUtils.toSQLString Derby fetch语句输出为limit
Derby支持fetch子句，不支持limit子句，SQLUtils.toSQLString会将fetch输出为limit
测试用例：

            String dbType = JdbcConstants.DERBY;
            String sql = "select * from mytest FETCH FIRST 200 ROWS ONLY";
            List<SQLStatement> statements = SQLUtils.parseStatements(sql, dbType);
            FormatOption DEFAULT_FORMAT_OPTION = new FormatOption(false, false, false);

            for(SQLStatement statement : statements) {
                String result = SQLUtils.toSQLString(statement, dbType, DEFAULT_FORMAT_OPTION);
                // 输出结果为: sql: select * from mytest limit 200
                System.out.println("sql: " + result);
            }

fudianchn:
## What

`SQLUtils.toSQLString(..., derby)` now renders the row-limiting clause as `FETCH FIRST n ROWS ONLY` (which Derby supports) instead of `LIMIT n` (which Derby does not support).

## Why

Derby uses the SQL standard `FETCH FIRST n ROWS ONLY` syntax and has no `LIMIT` keyword. druid parsed `FETCH FIRST n ROWS ONLY` correctly, but when re-serializing for Derby it emitted `LIMIT n`, producing SQL Derby rejects:

```
Input:  SELECT * FROM mytest FETCH FIRST 200 ROWS ONLY
Output: SELECT * FROM mytest LIMIT 200   ← invalid Derby SQL
```

## Root cause

`SQLASTOutputVisitor.printFetchFirst()` only emitted the `FETCH FIRST ... ROWS ONLY` form for `db2`, `oracle`, and `sqlserver`; every other dialect (including Derby) fell through to `printLimit()`, which writes `LIMIT`.

The `FETCH FIRST ... ROWS ONLY` row-limiting syntax is part of the SQL standard and shared by db2/derby/oracle/sqlserver, so Derby belongs in the same branch.

## How

Add `DbType.derby` to the dbType check in `printFetchFirst()` so Derby renders the fetch clause the same way db2/oracle/sqlserver already do. One line.

## Testing

- New test `DerbyFetchFirstTest#fetchFirstRendersAsFetchNotLimit` — fails before the fix (output `LIMIT 200`), passes after (output `FETCH FIRST 200 ROWS ONLY`).
- Regression: `*DB2SelectTest*`, `*FetchFirst*`, `*OracleSelectTest*` → `Tests run: 17, Failures: 0` (the existing db2/oracle FETCH paths are unaffected).

Before the fix:
```
SQLUtils.toSQLString(stmt, derby) → "SELECT *\nFROM mytest\nLIMIT 200"
```

After the fix:
```
SQLUtils.toSQLString(stmt, derby) → "SELECT *\nFROM mytest\nFETCH FIRST 200 ROWS ONLY"
```

Fixes #3612

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/visitor/SQLASTOutputVisitor.java
+++ b/core/src/main/java/com/alibaba/druid/sql/visitor/SQLASTOutputVisitor.java
@@ -2772,6 +2772,7 @@ public class SQLASTOutputVisitor extends SQLASTVisitorAdapter implements Paramet
         SQLExpr first = limit.getRowCount();
 
         if (DbType.db2 == dbType
+                || DbType.derby == dbType
                 || DbType.oracle == dbType
                 || DbType.sqlserver == dbType
         ) {
--- a/core/src/test/java/com/alibaba/druid/bvt/sql/derby/DerbyFetchFirstTest.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/sql/derby/DerbyFetchFirstTest.java
@@ -1,0 +1,25 @@
+package com.alibaba.druid.bvt.sql.derby;
+
+import com.alibaba.druid.sql.SQLUtils;
+import com.alibaba.druid.sql.ast.SQLStatement;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * Regression for issue #3612: Derby supports {@code FETCH FIRST n ROWS ONLY} and does NOT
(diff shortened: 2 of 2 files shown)


fudianchn:
Closing to reduce the Druid open-PR backlog I created by opening too many narrow parsing PRs at once. Derby FETCH FIRST row-limiting rendering is a narrow grammar addition; I'll revisit it later as part of a grouped pass rather than keep this standalone.

PR #6689, opened 2026-08-02
fix(oracle): allow dot in MyBatis placeholders ${a.b} / #{a.b}

### LINKED ISSUE #5172: com.alibaba.druid.sql.parser.ParserException: syntax errorpos
@Test
    public void test2() {
        String dbType = "oracle";
        String sql = "select * from table_a where b.b1 = ${a.basdf} ";
        SQLStatementParser parser = SQLParserUtils.createSQLStatementParser(sql, dbType);
        SQLSelectStatement stmt = (SQLSelectStatement) parser.parseStatement();
        stmt.getSelect().getQuery();
    }

fudianchn:
## What

Oracle's lexer now allows `.` inside MyBatis placeholders, so `${a.b}` / `#{a.b.c}` parse instead of raising `ParserException: syntax error`.

## Why

MyBatis parameter names commonly use dots (e.g. `${item.id}`). mysql/postgresql already accept this; the Oracle lexer's `scanVariable()` only allowed identifier characters and `:` inside `${...}`/`#{...}`, so the dot terminated the placeholder early and the leftover content broke parsing:

```
select * from table_a where b.b1 = ${a.basdf}
  → ParserException: syntax errorpos 38, line 1, column 36, token =
  at ...OracleLexer.scanVariable(OracleLexer.java:205)
```

## Root cause

`OracleLexer.scanVariable()` (the placeholder name loop) checked `!isIdentifierChar(ch) && ch != ':'` and broke on `.`. The base `Lexer.scanVariable()` does not have this restriction.

## How

Also accept `.` in that loop, mirroring how the base lexer handles dotted placeholder names. One line in the break condition.

## Testing

- New test `OracleMybatisPlaceholderTest` (4 cases), three fail before the fix with `syntax errorpos ... token =`, all pass after:
  - `mybatisPlaceholderWithDotParses` — the issue's SQL; also asserts round-trip.
  - `mybatisPlaceholderNestedDotParses` — `${a.b.c}`.
  - `hashPlaceholderWithDotParses` — same scan path covers `#{a.b}`.
  - `simplePlaceholderStillWorks` — regression guard: `${a}` (no dot) still works.
- Regression: `OracleLexerTest`, `OracleParameterTest` → `Tests run: 5, Failures: 0`.

Before:
```
SQLUtils.parseStatements("select * from table_a where b.b1 = ${a.basdf}", oracle)
  → ParserException: syntax errorpos 38 ... token =
```

After:
```
→ parses; round-trips to "SELECT *\nFROM table_a\nWHERE b.b1 = ${a.basdf}"
```

Fixes #5172

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleLexer.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/oracle/parser/OracleLexer.java
@@ -206,7 +206,11 @@ public class OracleLexer extends Lexer {
             for (; ; ) {
                 ch = charAt(++pos);
 
-                if (!isIdentifierChar(ch) && ch != ':') {
+                // Allow '.' only inside ${...} / #{...} MyBatis placeholders so dotted
+                // parameter names like ${a.b} are preserved. Plain :VAR bind variables
+                // (e.g. Oracle's :NEW."ID") must NOT consume the dot. See #5172.
+                boolean allowDot = mybatisFlag && ch == '.';
+                if (!isIdentifierChar(ch) && ch != ':' && !allowDot) {
                     break;
                 }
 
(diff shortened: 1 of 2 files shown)


fudianchn:
Closing to reduce the Druid open-PR backlog I created by opening too many narrow parsing PRs at once. dots in MyBatis placeholders ${a.b}/#{a.b} is a narrow grammar addition; I'll revisit it later as part of a grouped pass rather than keep this standalone.

PR #6694, opened 2026-08-02
fix(clickhouse): parse two-argument function-style CAST(x, 'Type')

### LINKED ISSUE #4421: clickhouse cast not support
such as : select CAST(data_day,'String') as dt 

Exception in thread "main" com.alibaba.druid.sql.parser.ParserException: syntax error, expect AS, actual , pos 17276, line 350, column 39, token ,

fudianchn:
## What

ClickHouse's function-style two-argument `CAST(x, 'TypeName')` now parses, instead of failing with `expect AS, actual ,`.

## Why

ClickHouse supports both `CAST(x AS type)` and `CAST(x, 'TypeName')`. druid only parsed the AS form, so:

```
SELECT CAST(data_day, 'String') AS dt
  → ParserException: syntax error, expect AS, actual , pos 17276 ... token ,
```

## Root cause

The base `SQLExprParser.parseCast()` only accepts `CAST(expr AS type)`. `CKExprParser` did not override it.

## How

Override `parseCast()` in `CKExprParser`: after parsing the first expression, if the next token is `,`, consume it and parse the quoted type name; the type string (e.g. `'String'`, `'Int64'`, `'Nullable(Int64)'`) is fed through the regular data-type parser to produce the target `SQLDataType`. Otherwise fall through to the standard `CAST(expr AS type)` handling.

Round-trip emits the `CAST(x AS type)` form, which ClickHouse also accepts.

## Testing

- New test `CKCastTwoArgTest` (3 cases), two fail before the fix, all pass after:
  - `castTwoArgWithStringType` — the issue's SQL.
  - `castTwoArgWithNumericType` — `CAST(value, 'Int64')`.
  - `castAsFormStillWorks` — regression guard for `CAST(x AS type)`.
- Regression: `CK_*` → `Tests run: 9, Failures: 0`.

Before:
```
SELECT CAST(data_day, 'String') AS dt  →  expect AS, actual ,
```

After:
```
→ parses successfully
```

Fixes #4421

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/clickhouse/parser/CKExprParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/clickhouse/parser/CKExprParser.java
@@ -21,6 +21,7 @@ import com.alibaba.druid.sql.ast.SQLExpr;
 import com.alibaba.druid.sql.ast.SQLName;
 import com.alibaba.druid.sql.ast.SQLStructDataType;
 import com.alibaba.druid.sql.ast.expr.SQLArrayExpr;
+import com.alibaba.druid.sql.ast.expr.SQLCastExpr;
 import com.alibaba.druid.sql.ast.expr.SQLCharExpr;
 import com.alibaba.druid.sql.ast.expr.SQLIdentifierExpr;
 import com.alibaba.druid.sql.ast.statement.SQLColumnDefinition;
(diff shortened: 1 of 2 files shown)


fudianchn:
Closing to reduce the Druid open-PR backlog I created by opening too many narrow parsing PRs at once. ClickHouse two-argument function-style CAST is a narrow grammar addition; I'll revisit it later as part of a grouped pass rather than keep this standalone.

PR #6695, opened 2026-08-02
fix(h2): register LISTAGG as aggregate for WITHIN GROUP support

### LINKED ISSUE #4301: H2数据库语法检测错误
`com.alibaba.druid.sql.parser.ParserException: syntax error, error in :'thin group(order by length(fdn) asc', expect BY, actual LPAREN pos 55, line 1, column 55, token LPAREN`

H2有oralce模式支持 listagg within group 语法

版本：1.1.24

fudianchn:
## What

H2 now parses `LISTAGG(...) WITHIN GROUP (ORDER BY ...)` (H2's Oracle-compatibility mode), instead of failing with `expect BY, actual (`.

## Why

H2 supports Oracle-compatibility mode including the `LISTAGG` ordered-set aggregate. druid's H2 parser did not register `LISTAGG` as an aggregate function, so it was parsed as a plain method call and the trailing `WITHIN GROUP (ORDER BY ...)` clause broke parsing:

```
SELECT listagg(x, ',') WITHIN GROUP (ORDER BY length(fdn) ASC) FROM t
  → ParserException: syntax error ... expect BY, actual (, pos 37 ... token (
```

The same SQL parses fine on the oracle dialect, whose aggregate-function list already includes `LISTAGG`.

## Root cause

`H2ExprParser`'s `AGGREGATE_FUNCTIONS` array omitted `LISTAGG`, so `isAggregateFunction("LISTAGG")` returned false and the WITHIN GROUP aggregate path (`parseAggregateExpr` → `parseAggregateExprRest`) was never taken.

## How

Add `LISTAGG` to `H2ExprParser.AGGREGATE_FUNCTIONS`, matching the Oracle parser. The base `parseAggregateExprRest` already handles `WITHIN GROUP`.

## Testing

- New test `H2ListaggWithinGroupTest` (2 cases), both fail before the fix, both pass after:
  - `listaggWithinGroup` — the issue's SQL.
  - `listaggWithinGroupRoundTrips` — asserts the statement round-trips.
- Regression: `H2_*` → `Tests run: 3, Failures: 0`.

Before:
```
SELECT listagg(x, ',') WITHIN GROUP (ORDER BY length(fdn) ASC) FROM t  →  expect BY, actual (
```

After:
```
→ parses; round-trips with WITHIN GROUP preserved.
```

Fixes #4301

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/h2/parser/H2ExprParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/h2/parser/H2ExprParser.java
@@ -29,7 +29,7 @@ public class H2ExprParser extends SQLExprParser {
     private static final long[] AGGREGATE_FUNCTIONS_CODES;
 
     static {
-        String[] strings = {"AVG", "COUNT", "MAX", "MIN", "STDDEV", "SUM", "ROW_NUMBER",
+        String[] strings = {"AVG", "COUNT", "LISTAGG", "MAX", "MIN", "STDDEV", "SUM", "ROW_NUMBER",
                 "ROWNUMBER"};
         AGGREGATE_FUNCTIONS_CODES = FnvHash.fnv1a_64_lower(strings, true);
         AGGREGATE_FUNCTIONS = new String[AGGREGATE_FUNCTIONS_CODES.length];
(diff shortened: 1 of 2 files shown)


fudianchn:
Closing to reduce the Druid open-PR backlog I created by opening too many narrow parsing PRs at once. H2 LISTAGG aggregate registration is a narrow grammar addition; I'll revisit it later as part of a grouped pass rather than keep this standalone.

PR #6697, opened 2026-08-02
fix(sql): support DROP FOREIGN TABLE parsing

### LINKED ISSUE #6180: [BUG] holo DROP FOREIGN TABLE
### Database Type

hologres

### Database Version

Hologres V1.3

### Druid Version

1.2.23

### JDK Version

Open JDK 8

### Error SQL

DROP FOREIGN TABLE odps_user_test;

### Testcase Code

_No response_

### Stacktrace Info

com.alibaba.druid.sql.parser.ParserException: TODO pos 12, line 1, column 6, token FOREIGN
	at com.alibaba.druid.sql.parser.SQLStatementParser.parseDrop(SQLStatementParser.java:920) ~[druid-1.2.24.1.jar:na]
	at com.alibaba.druid.sql.parser.SQLStatementParser.parseStatementList(SQLStatementParser.java:427) ~[druid-1.2.24.1.jar:na]
	at com.alibaba.druid.sql.parser.SQLStatementParser.parseStatementList(SQLStatementParser.java:108) ~[druid-1.2.24.1.jar:na]

### Error Info

TODO pos 12, line 1, column 6, token FOREIGN

fudianchn:
## What
Add `DROP FOREIGN TABLE` parsing support. PostgreSQL's `FOREIGN TABLE` (foreign-data wrapper) was not recognised by the DROP statement parser.

## Why
`SQLUtils.parseStatements("DROP FOREIGN TABLE odps_user_test", "postgresql")` raises:

```
ParserException: unsupported DROP target 'FOREIGN'. pos 12, line 1, column 13
```

`FOREIGN TABLE` is standard PostgreSQL DDL (`DROP FOREIGN TABLE [ IF EXISTS ] name [, ...] [ CASCADE | RESTRICT ]`), so this breaks parsing of any PG script that contains it. Issue #6180.

## Root cause
`SQLStatementParser.parseDrop()`'s switch has no `case FOREIGN`, and the default branch only special-cases identifier keywords (EXTERNAL, ROLE, …); the `FOREIGN` token falls straight to `throw new ParserException("unsupported DROP target 'FOREIGN'...")`. `parseDropTable()` likewise never consumes a FOREIGN modifier.

## How
- `SQLDropTableStatement`: add a dedicated `foreign` boolean flag (mirrors the existing `external` flag), wired into getter/setter/clone/equals/hashCode.
- `parseDrop()`: add `case FOREIGN` → `parseDropTable(false)` (parallel to `case TABLE`).
- `parseDropTable()`: consume the FOREIGN token and set the flag.
- `SQLASTOutputVisitor.visit(SQLDropTableStatement)`: render `FOREIGN` when the flag is set.

A dedicated `foreign` flag (rather than reusing `external`) is deliberate: PostgreSQL has no `EXTERNAL TABLE` syntax, so reusing `external` would render round-tripped SQL as `DROP EXTERNAL TABLE ...`, which is invalid for PostgreSQL. The new flag keeps the round-trip lossless (`DROP FOREIGN TABLE x` → `DROP FOREIGN TABLE x`).

## Testing
- New `Issue6180Test`: plain `DROP FOREIGN TABLE` round-trip + `isForeign()` assertion; `IF EXISTS` + multi-table + `CASCADE`; regression guard that ordinary `DROP TABLE` does not set the flag. (`RESTRICT` is omitted because PG's DROP parser doesn't accept it on any table kind — a separate, pre-existing limitation out of scope for this issue.)
- `./mvnw -pl core test`: **4376 tests, 0 failures, 0 errors** — full core regression green.

Fixes #6180

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/ast/statement/SQLDropTableStatement.java
+++ b/core/src/main/java/com/alibaba/druid/sql/ast/statement/SQLDropTableStatement.java
@@ -37,6 +37,7 @@ public class SQLDropTableStatement extends SQLStatementImpl implements SQLDropSt
     protected boolean ifExists;
     private boolean temporary;
     private boolean external;
+    private boolean foreign;
     private boolean isDropPartition;
     private SQLExpr where;
 
@@ -152,6 +153,14 @@ public class SQLDropTableStatement extends SQLStatementImpl implements SQLDropSt
         this.external = external;
     }
 
+    public boolean isForeign() {
+        return foreign;
+    }
+
+    public void setForeign(boolean foreign) {
+        this.foreign = foreign;
+    }
+
     public List<SQLCommentHint> getHints() {
         return hints;
     }
@@ -203,6 +212,7 @@ public class SQLDropTableStatement extends SQLStatementImpl implements SQLDropSt
         x.ifExists = ifExists;
         x.temporary = temporary;
         x.external = external;
+        x.foreign = foreign;
         x.isDropPartition = isDropPartition;
         if (where != null) {
             x.where = where.clone();
@@ -231,6 +241,7 @@ public class SQLDropTableStatement extends SQLStatementImpl implements SQLDropSt
                 && ifExists == that.ifExists
                 && temporary == that.temporary
                 && external == that.external
+                && foreign == that.foreign
                 && isDropPartition == that.isDropPartition
                 && Objects.equals(hints, that.hints)
                 && Objects.equals(tableSources, that.tableSources)
@@ -247,6 +258,7 @@ public class SQLDropTableStatement extends SQLStatementImpl implements SQLDropSt
         result = 31 * result + Boolean.hashCode(ifExists);
         result = 31 * result + Boolean.hashCode(temporary);
         result = 31 * result + Boolean.hashCode(external);
+        result = 31 * result + Boolean.hashCode(foreign);
         result = 31 * result + Boolean.hashCode(isDropPartition);
         result = 31 * result + Objects.hashCode(where);
         return result;
--- a/core/src/main/java/com/alibaba/druid/sql/parser/SQLStatementParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/parser/SQLStatementParser.java
@@ -905,6 +905,14 @@ public class SQLStatementParser extends SQLParser {
             case SEQUENCE:
                 stmt = parseDropSequence(false);
                 break;
+            case FOREIGN: {
+                SQLDropTableStatement dropTable = parseDropTable(false);
+                if (hints != null) {
+                    dropTable.setHints(hints);
+                }
+                stmt = dropTable;
+                break;
+            }
             case TABLE: {
                 SQLDropTableStatement dropTable = parseDropTable(false);
                 if (temporary) {
@@ -3328,6 +3336,11 @@ public class SQLStatementParser extends SQLParser {
         if (lexer.identifierEquals(FnvHash.Constants.EXTERNAL)) {
             lexer.nextToken();
             stmt.setExternal(true);
+        }
+
+        if (lexer.token == Token.FOREIGN) {
+            lexer.nextToken();
+            stmt.setForeign(true);
         }
 
         if (lexer.token == TABLE) {
--- a/core/src/main/java/com/alibaba/druid/sql/parser/SQLStatementParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/parser/SQLStatementParser.java
@@ -907,6 +907,9 @@ public class SQLStatementParser extends SQLParser {
                 break;
             case FOREIGN: {
                 SQLDropTableStatement dropTable = parseDropTable(false);
+                if (temporary) {
+                    dropTable.setTemporary(true);
+                }
                 if (hints != null) {
                     dropTable.setHints(hints);
                 }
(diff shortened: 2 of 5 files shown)


core/src/main/java/com/alibaba/druid/sql/parser/SQLStatementParser.java
```diff
@@ -905,6 +905,14 @@ public SQLStatement parseDrop() {
             case SEQUENCE:
                 stmt = parseDropSequence(false);
                 break;
+            case FOREIGN: {
+                SQLDropTableStatement dropTable = parseDropTable(false);
+                if (hints != null) {
+                    dropTable.setHints(hints);
+                }
```
wenshao, line 915: **[Suggestion]** `case FOREIGN:` omits propagation of the `temporary` flag, unlike the adjacent `case TABLE:` which does `if (temporary) { dropTable.setTemporary(true); }`. — Concrete cost: `DROP TEMPORARY FOREIGN TABLE t1` parses without error but the AST silently drops `TEMPORARY` (probe-confirmed: `isTemporary=false`, round-trip outputs `DROP FOREIGN TABLE t1`). No real dialect supports this combination, so impact is theoretical, but the parser accepts invalid input silently rather than rejecting or preserving it.

```suggestion
            case FOREIGN: {
                SQLDropTableStatement dropTable = parseDropTable(false);
                if (temporary) {
                    dropTable.setTemporary(true);
                }
                if (hints != null) {
                    dropTable.setHints(hints);
                }
```

_— qwen3.8-max-preview via Qwen Code /review (v0.21.3)_

core/src/test/java/com/alibaba/druid/bvt/sql/postgresql/issues/Issue6180Test.java
```diff
@@ -0,0 +1,55 @@
...
+ * PostgreSQL 解析 DROP FOREIGN TABLE 报错的问题。
+ *
+ * <p>FOREIGN TABLE 是 PostgreSQL 外部表（FDW）的独立语法，druid 原本在 DROP 语句中
+ * 未识别 FOREIGN 关键字，抛出 "unsupported DROP target 'FOREIGN'"。
+ *
+ * @author fudianchn
+ * @see <a href="https://github.com/alibaba/druid/issues/6180">Issue #6180</a>
+ * @see <a href="https://www.postgresql.org/docs/current/sql-droptable.html">DROP TABLE</a>
+ */
+public class Issue6180Test {
```
wenshao: **[Suggestion]** Test class name `Issue6180Test` breaks the directory naming convention — all 38 sibling files in `postgresql/issues/` use `Issue<NNNN>` without a `Test` suffix (e.g. `Issue6146.java`, `Issue6637.java`). — Concrete cost: a script enumerating issue-regression tests by the directory's own pattern (`Issue\d+\.java`) would miss this file.

```suggestion
public class Issue6180 {
```

_— qwen3.8-max-preview via Qwen Code /review (v0.21.3)_

core/src/main/java/com/alibaba/druid/sql/ast/statement/SQLDropTableStatement.java
```diff
@@ -203,6 +212,7 @@ protected void cloneTo(SQLDropTableStatement x) {
         x.ifExists = ifExists;
         x.temporary = temporary;
         x.external = external;
+        x.foreign = foreign;
         x.isDropPartition = isDropPartition;
```
wenshao, line 216: **[Suggestion]** The new `foreign` field's propagation in `cloneTo`, `equals`, and `hashCode` has no test coverage. — Concrete cost: if a future refactor accidentally drops `x.foreign = foreign` from `cloneTo`, every existing test stays green, but clone-based SQL rewriting paths would silently regenerate `DROP TABLE` from `DROP FOREIGN TABLE`, producing invalid PostgreSQL that fails at execution rather than parse time.

Consider adding to the test:
```java
SQLDropTableStatement clone = drop.clone();
assertTrue(clone.isForeign());
assertEquals(sql, clone.toString());
```

_— qwen3.8-max-preview via Qwen Code /review (v0.21.3)_

wenshao (commented):

_— qwen3.8-max-preview via Qwen Code /review (v0.21.3)_

core/src/main/java/com/alibaba/druid/sql/parser/SQLStatementParser.java
fudianchn: Fixed in d366e14 — the FOREIGN branch now sets `dropTable.setTemporary(true)` when `temporary` holds, symmetric with the TABLE case right above it.

core/src/test/java/com/alibaba/druid/bvt/sql/postgresql/issues/Issue6180Test.java
fudianchn: Renamed in d366e14 — `Issue6180Test` is now `Issue6180` (file + class), matching the `Issue<NNNN>` convention in this directory. druid surefire uses `<include>**/bvt/**/*.java</include>` (core/pom.xml), so the Test suffix was never required; confirmed it still runs (`Tests run: 4`).

core/src/main/java/com/alibaba/druid/sql/ast/statement/SQLDropTableStatement.java
fudianchn: Added in d366e14 — `clonePreservesForeignFlag` asserts `clone().isForeign()`, `!isExternal()`, and a round-trip `toString()` match. Confirmed the `foreign` field is already propagated in `cloneTo` / `equals` / `hashCode`, so this is a regression guard rather than a bug fix.

fudianchn:
Closing for now to cut the backlog I created by opening too many Druid PRs at once. DROP FOREIGN TABLE parsing is a fuller change; I'll re-file it once the open count is down so it gets focused review.

PR #6699, opened 2026-08-02
fix(hive): allow COMMENT between PARTITIONED BY and LOCATION in CREATE TABLE

### LINKED ISSUE #6071: [BUG] 解析hive时如果  PARTITIONED BY  和 LOCATION 关键字之间有COMMENT解析失败
### Database Type

hivi

### Database Version

无

### Druid Version

1.2.23

### JDK Version

jdk8

### Error SQL

CREATE TABLE IF NOT EXISTS test.dwr_whs_invt_dtl(
bat_no  string COMMENT '批号',
.................................................
)
USING hudi
PARTITIONED BY (acct_prd string comment '会计期') 
COMMENT '明细表'
LOCATION 'obs://xxxxxxxxxxxxxxxxxx'
TBLPROPERTIES (
  type = 'mor', 
  primaryKey = 'acct_prd' ,
  preCombineField = 'acct_prd'
);

### Testcase Code

_No response_

### Stacktrace Info

_No response_

### Error Info

_No response_

fudianchn:
## What
Hive `CREATE TABLE` now accepts a table-level `COMMENT` placed between `PARTITIONED BY` and `LOCATION`.

## Why
```sql
CREATE TABLE IF NOT EXISTS test.t (
  bat_no string COMMENT '批号'
)
PARTITIONED BY (acct_prd string comment '会计期')
COMMENT '明细表'
LOCATION 'obs://bucket/path'
TBLPROPERTIES ('type'='mor')
```
raised `not supported ... token IDENTIFIER LOCATION`. Issue #6071.

## Root cause
`HiveCreateTableParser.parseCreateTableRest` is a linear chain of `if` guards that each recognise a clause once. `COMMENT` was only recognised in two positions — before `PARTITIONED BY` (line 153) and after `LOCATION`/`TBLPROPERTIES` (line 337). When `COMMENT` sat between `PARTITIONED BY` and `LOCATION`, it slipped through every intermediate guard and was consumed by the trailing `COMMENT` branch — after which the `LOCATION` token had no matching branch left (the only `LOCATION` guard sits before that point) and the parser failed.

## How
Recognise `COMMENT` once more, immediately before the `LOCATION` branch, so the clause is consumed in the order the user wrote it. The pre-existing COMMENT branches (before PARTITIONED BY, and after TBLPROPERTIES) are untouched, so previously supported orderings keep working.

## Testing
- New `Issue6071Test`: the issue's `PARTITIONED BY ... COMMENT ... LOCATION ...` ordering now parses, capturing both COMMENT and LOCATION; a regression guard checks the previously supported `LOCATION ... COMMENT` ordering still works.
- `./mvnw -pl core test`: full core suite green.

Fixes #6071

### DIFF
--- a/core/src/main/java/com/alibaba/druid/sql/dialect/hive/parser/HiveCreateTableParser.java
+++ b/core/src/main/java/com/alibaba/druid/sql/dialect/hive/parser/HiveCreateTableParser.java
@@ -292,6 +292,10 @@ public class HiveCreateTableParser extends SQLCreateTableParser {
             stmt.setRbracketUse(true);
             lexer.nextToken();
         }
+        if (lexer.nextIf(Token.COMMENT)) {
+            SQLExpr comment = this.exprParser.expr();
+            stmt.setComment(comment);
+        }
         if (lexer.identifierEquals(FnvHash.Constants.LOCATION)) {
             lexer.nextToken();
             SQLExpr location = this.exprParser.primary();
--- a/core/src/test/java/com/alibaba/druid/bvt/sql/hive/issues/Issue6071Test.java
+++ b/core/src/test/java/com/alibaba/druid/bvt/sql/hive/issues/Issue6071Test.java
@@ -1,0 +1,47 @@
+package com.alibaba.druid.bvt.sql.hive.issues;
+
+import com.alibaba.druid.sql.SQLUtils;
+import com.alibaba.druid.sql.ast.SQLStatement;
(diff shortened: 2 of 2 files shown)


fudianchn:
Closing to reduce the Druid open-PR backlog I created by opening too many narrow parsing PRs at once. Hive COMMENT between PARTITIONED BY and LOCATION is a narrow grammar addition; I'll revisit it later as part of a grouped pass rather than keep this standalone.

PR #6717, opened 2026-09-08
fix: reject trailing garbage after Oracle UPDATE statement

### LINKED ISSUE #6702: [BUG] 对于oracle的解析过于宽松
### Database Type

Oracle

### Database Version

任意版本

### Druid Version

1.2.28

### JDK Version

JDK21

### Error SQL

String sql = "UPDATE test SET value = 1 WHER id = 1";

List<SQLStatement> statements =
        SQLUtils.parseStatements(sql, DbType.oracle);

当前返回：

OracleUpdateStatement: UPDATE test SET value = 1
SQLExprStatement: WHER
SQLExprStatement: id = 1

### Testcase Code

_No response_

### Stacktrace Info

_No response_

### Error Info

_No response_

wylovelyi:
## Summary
- Oracle's `UPDATE` parser (`OracleUpdateParser#parseUpdateStatement`) did not validate the end token. A typo'd clause such as `WHER` instead of `WHERE` was silently swallowed as an extra statement instead of raising an error.
- Every other dialect (mysql, postgresql, db2, sqlserver, odps, dm, clickhouse, ...) already rejects the same input through `SQLStatementParser#checkEndToken`; Oracle was the only outlier.

## Root cause
`OracleUpdateParser.parseUpdateStatement()` returned the update statement without an end-token check. The outer `OracleStatementParser.parseStatementList()` IDENTIFIER fallback branch then wrapped the trailing garbage (`WHER id = 1`) into a spurious `SQLExprStatement`, so the parser returned multiple statements for a single invalid `UPDATE`.

## Fix
Add an `EOF`/`SEMI` end-token check right before `return update;` in `OracleUpdateParser#parseUpdateStatement`, throwing `ParserException("not supported. " + lexer.info())` — identical to the error format emitted by the base `checkEndToken()`. This aligns Oracle with the other dialects without touching the `IDENTIFIER` fallback (which is still needed for legitimate PL/SQL function calls).

## Test
Added `Issue6702` (in `core/src/test/java/com/alibaba/druid/bvt/sql/oracle/issues/`):
- `UPDATE test SET value = 1 WHER id = 1` → now throws `ParserException` (previously silently split into multiple statements)
- `UPDATE test SET value = 1 XYZ id = 1` → throws `ParserException`
- `UPDATE test SET value = 1 WHERE id = 1` → still parses as a single statement
- MySQL baseline already rejects the same typo'd SQL

Regression: the full `com.alibaba.druid.bvt.sql.oracle.**` test package was run — **157 tests, all green**.

Fixes #6702


wylovelyi:
I have signed the CLA.

wylovelyi:
recheck

Issue #6722, opened 2026-09-18
[BUG] Hive SQLUtils.format duplicates and merges comments before SELECT items


WCY-dt:
## Description

When formatting Hive SQL, `SQLUtils.format` does not preserve line comments before SELECT items correctly.

In Druid 1.2.8, these comments are removed completely. In the latest release, Druid 1.2.28, the comments are retained partially but are duplicated, concatenated, and associated with the wrong SELECT items.

## Environment

- Database Type: Hive
- Database Version: N/A — the problem occurs in the SQL formatter without connecting to a database
- Druid Version: 1.2.28
- JDK Version: OpenJDK 17.0.19

## Minimal SQL

```sql
SELECT
    -- first column
    a,
    -- second column
    b
FROM demo_table
```

## Minimal Reproduction

```java
import com.alibaba.druid.DbType;
import com.alibaba.druid.sql.SQLUtils;

public class DruidHiveCommentFormatTest {

    public static void main(String[] args) {
        String sql =
                "SELECT\n"
              + "    -- first column\n"
              + "    a,\n"
              + "    -- second column\n"
              + "    b\n"
              + "FROM demo_table";

        String formatted = SQLUtils.format(sql, DbType.hive);
        System.out.println(formatted);
    }
}
```

## Actual Output with Druid 1.2.28

```sql
SELECT -- first column
    a -- first column-- second column
    
    , b
FROM demo_table
```

The output has the following problems:

1. `-- first column` is emitted twice.
2. `-- first column` and `-- second column` are concatenated.
3. The comments are no longer associated with the correct SELECT items.
4. The formatter changes the comment structure and meaning of the original SQL.

## Expected Output

The exact indentation is not important, but each comment should be emitted exactly once, in the original order, and remain associated with the corresponding SELECT item.

For example:

```sql
SELECT
    -- first column
    a,
    -- second column
    b
FROM demo_table
```

The formatter should guarantee that:

1. no comment is lost;
2. no comment is duplicated;
3. comments are not concatenated;
4. comment order is preserved;
5. each comment remains associated with the correct SQL element.

## Behavior in Druid 1.2.8

Running the same reproduction with Druid 1.2.8 produces:

```sql
SELECT a, b
FROM demo_table
```

In 1.2.8, both comments are removed completely.

Upgrading from 1.2.8 to 1.2.28 therefore does not resolve the problem. The behavior changes from dropping comments to duplicating and merging them.

## Additional Analysis

[...]

PR #6723, opened 2026-09-19
Fix leading SELECT item comments when formatting SQL

jakezwang:
Addresses the leading SELECT-item comment example in #6722.

Consume pending comments before parsing each SELECT item so the same comment is not attached both before and after the item. Print a later item's leading comments after the comma, retaining the necessary line break in compact output as well.

Adds Hive AST and formatting regressions for leading line/block comments, trailing comments and pretty/compact roundtrips. Updates two existing ODPS/PostgreSQL snapshots that contained duplicated or misplaced SELECT comments; their input SQL is unchanged.

Validation:
- Full JDK 8 reactor with Javadoc: 4,382 tests, one skipped.
- Full JDK 17 reactor with coverage: 4,394 tests, one skipped.
- Checkstyle and Hive/PostgreSQL formatting roundtrip checks.

Scope: this change handles comments before SELECT items. It does not fix comments between expressions and aliases, mixed trailing/leading comments, or trailing block comments. The four additional cases in review reproduce on both the base and this PR head; they remain separate parser/formatter issues.


WCY-dt (commented):
Thanks again for fixing the reproduction in #6722. I expanded the edge-case testing and would like to qualify my earlier review: the original example is fixed, but the formatter still does not satisfy all of the comment-preservation guarantees described in the issue.

I checked PR head `51f9240` against base `fa8dc99`. **All four cases below also fail on the base; I am not reporting them as regressions introduced by this PR.** They are remaining coverage gaps, and it would be helpful to decide which belong in this fix and which should be tracked separately.

Using `SQLUtils.format(sql, DbType.hive)`:

1. **A comment between an expression and its alias can swallow the alias and the next column.**

   Input:
   ```sql
   SELECT a + 1 -- explanation
   AS x, b FROM t
   ```
   Output on both commits:
   ```sql
   SELECT a + 1 -- explanation AS x, b
   FROM t
   ```
   Reparsing the output gives one SELECT item instead of two, and loses alias `x`. An assertion on the number of comments or formatting stability does not catch this; an assertion on the projection/alias does.

2. **A trailing comment followed by a leading comment is not split between the two items.**
   ```sql
   SELECT a, -- describes a
   -- describes b
   b FROM t
   ```
   The parser puts both comments in `b.getBeforeCommentsDirect()` and leaves `a.getAfterCommentsDirect()` empty. The existing trailing-comment test covers only a single comment; this two-comment variant would make a useful additional AST regression.

3. **A line comment followed by a block comment is merged into one line comment.**
   ```sql
   SELECT a -- trailing line
   /* trailing block */, b FROM t
   ```
   The output contains `a -- trailing line/* trailing block */`. Both texts survive, but the two original comment tokens become one. Comparing the actual before/after comment lists catches this, whereas counting each text once does not.

4. **A trailing block comment is lost.**
   ```sql
   SELECT a /* describes a */, b FROM t
   ```
   The output is `SELECT a, b FROM t` (with the normal pretty-format newline).

The three new PR tests still pass. I added independent expectation tests for the four cases above: all four fail on both commits. The expanded generated checks cover ten dialects, pretty/compact and upper/lowercase output, comment positions inside expressions and around aliases, nested queries, CTEs, UNIONs, LF/CRLF/CR, and mixed/repeated comments. JaCoCo exercised all 10 modified executable lines and all 12 branch outcomes on those lines, which also illustrates why executing the changed code alone is not enough to establish comment preservation.

My conclusion is narrower than my initial note: this PR fixes the original leading-item example, but it should not be taken as confirmation that arbitrary comment placement is safe. I have not established a new regression attributable to this diff. I appreciate the focused fix; the examples above are intended to help strengthen the tests and clarify the remaining scope, not to attribute these existing problems to your contribution. I have not rerun the full reactor; the broader existing-test selection still has the previously noted, unrelated `CommentTest.test_2` baseline failure.

jakezwang:
Confirmed all four examples on both `fa8dc99` and `51f9240`, including the lost alias/projection in the first case. I have narrowed the description to leading SELECT-item comments and removed the automatic issue-closing reference. The expression/alias and trailing/mixed-comment cases remain outside this patch; I suggest tracking those separately rather than extending this fix across additional comment-attachment paths.

